PR #3, opened 2017-01-18
Fixed Readme and some code.

homelessnessbo:
Hi! I've checkouted master branch of your package and got some problems there. There is fix that works for me, take a look pls. 

There is errors I got without that fix: 
```
vendor/github.com/crosbymichael/cgroups/cpuset.go:49: cannot use resources.CPU.Cpus (type string) as type *string in field value
vendor/github.com/crosbymichael/cgroups/cpuset.go:53: cannot use resources.CPU.Mems (type string) as type *string in field value
```
P.S. I am using go1.7.4 linux/amd64.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ uses the v1 implementation of cgroups.
 
 
 ```go
-shares := 100
-control, err := cgroups.New(cgroups.V1, cgroups.StaticPath("/test"), &specs.Resources{
+shares := uint64(100)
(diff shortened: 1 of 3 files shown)


crosbymichael:
LGTM

crosbymichael:
I'll vendor in the runtime spec whenever it gets frozen for this package.  So for right now we have to deal with broken builds if they change types.

Thanks for the fix.

PR #13, opened 2017-06-26
Add a vendor directory

justincormack:
CI is currently failing as go-metrics no longer builds with current Prometheus master.
Add a vendor directory that matches the required subset of `containerd/containerd` to fix the build.


justincormack:
ping @crosbymichael

dqminh:
Usually i dont really like vendoring in a library though. Maybe we should fix the build in go-metrics instead ?

justincormack:
The problem is that Prometheus changed in an incompatible way, and they have not done another release since, we are using 0.8, so it seems that vendoring is the only sane thing to do for now.

crosbymichael:
I think I did this initially and we run into some issues when we add this as a vendor to other projects that are using the vendor directory.  Maybe we could just checkout the right prom commit/tag in the `.travis.yml` file instead?

dqminh:
> I think I did this initially and we run into some issues when we add this as a vendor to other projects that are using the vendor directory.

Yah afaik there are vendoring programs that dont strip out the libraries' vendor dir by default, which can causes problems.

PR #15, opened 2017-06-26
Make Travis use Prometheus 0.8.0

justincormack:
Otherwise build will fail with Prometheus master

Also had to pin runtime spec until we update to the next rc.

replaces #13


crosbymichael:
LGTM

PR #16, opened 2017-06-27
Missing error return if hugetlb support missing

estesp:
Without returning an error, the `defaults()` subsystem builder ignores
that the hugetlb subsystem is nil and causes a panic later when any
attempt to walk and use the subsystem slice happens.

### DIFF
--- a/hugetlb.go
+++ b/hugetlb.go
@@ -13,7 +13,7 @@ import (
 func NewHugetlb(root string) (*hugetlbController, error) {
 	sizes, err := hugePageSizes()
 	if err != nil {
-		return nil, nil
+		return nil, err
 	}
 
 	return &hugetlbController{
--- a/utils.go
+++ b/utils.go
@@ -49,12 +49,11 @@ func runningInUserNS() bool {
 // defaults returns all known groups
 func defaults(root string) ([]Subsystem, error) {
 	h, err := NewHugetlb(root)
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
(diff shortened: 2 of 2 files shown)


hugetlb.go
```diff
@@ -13,7 +13,7 @@ import (
 func NewHugetlb(root string) (*hugetlbController, error) {
 	sizes, err := hugePageSizes()
 	if err != nil {
-		return nil, nil
+		return nil, err
 	}
```
dqminh: this is not a critical error in runc too, we just let sizes be an empty slice and continue.

We cant get any stats in that case though but the system is probably not in the correct state either.

estesp: Hmm..I meant to make this non-critical as well, but now I see that the `defaults()` setup errors out instead of just not including hugetlb subsystem. So at this point it is panic or failure.. let me try and update the PR

utils.go
```diff
@@ -71,6 +66,11 @@ func defaults(root string) ([]Subsystem, error) {
 	if !isUserNS {
 		s = append(s, NewDevices(root))
 	}
+	// add the hugetlb cgroup if test of support doesn't error out
+	h, hErr := NewHugetlb(root)
+	if hErr == nil {
```
crosbymichael: I think you can maybe check for a `os.IsNotExist` error.  We can skip if it is an notexists error and return it if its anything else?

estesp:
@crosbymichael makes sense--updated

crosbymichael:
LGTM

PR #17, opened 2017-06-27
Remove prometheus metrics from cgroups

crosbymichael:
Removing the prom code from cgroups allows more flexibility for consumers to structure metrics how they need.


stevvooe:
@crosbymichael Presumably, we are moving up the stack to assemble these metrics?

crosbymichael:
@stevvooe ya, moving it into containerd

memory.go
```diff
@@ -63,7 +63,7 @@ func (m *memoryController) Update(path string, resources *specs.LinuxResources)
 	if resources.Memory == nil {
 		return nil
 	}
-	g := func(v *uint64) bool {
+	g := func(v *int64) bool {
```
justincormack: These are already merged - rebase issue?

crosbymichael: No they are not, not in this package

justincormack: Ah yes, sorry!

PR #24, opened 2017-08-11
fix exp delay in remove

kung-foo:
Fixes a minor bug in the exponential delay when deleting a cgroup.

Signed-off-by: Jonathan Camp <github@jonathan.camp>


kung-foo:
Just doing my part to make Docker a little bit slower ;)

Issue #5, opened 2017-01-26
bug: memory.memsw.usage_in_bytes : no such file or directory


homelessnessbo:
Hi there! 

I have a problems with [Stat() ](https://github.com/crosbymichael/cgroups/blob/master/cgroup.go#L158)function in cgroups package:

Code example(approximately): 
```golang
control, err = cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupName), &specs.LinuxResources {})
err := control.Add(cgroups.Process{Pid:int(pid)})
stats, err := control.Stat()
usage := stats.Cpu.Usage.Total
fmt.Printf("\n%v", usage)
```

It's just a part of a code, but its quite simple for seeing the logic I go through.

But when I try to run it:
* Process that I expect to be moved into `cgroupName` is moved, it's ok.
* My code falls into panic with `/sys/fs/cgroup/memory/somecgroupname/memory.memsw.usage_in_bytes:  no such file or directory`

So, what have I found: `memory.memsw.usage_in_bytes` file isnt always present in system because its optionaly cgroups feature that could be activated by [passing `swapaccount=1`](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1348688/comments/7) into kernel (actualy, I'm using Ubunt/Debian systems in my environment, so I have the same situation as I linked you to)

Additional links about kernel parameters that has impackt to file existing factor:

http://cateee.net/lkddb/web-lkddb/MEMCG_SWAP.html
http://cateee.net/lkddb/web-lkddb/MEMCG_SWAP_ENABLED.html.

So, it looks like we need some check about if the file exists or it doesnt to prevent that kind of panic.
May be I will try to figure it out by myself and create PR then, but if you have enough time and see quite easy way to do it, it would be nice.

Anyway, thank you for opensourcing that again!

crosbymichael:
First is the panic coming from your code or this cgroups package?  If its coming from the package can you post the stack and I can fix it or you can submit a PR if you have time.  

To fix this fast you should be able to pass an error handler for the stats function.  This is one already in this package to ignore isnotexist errors.

```go
control.Stats(cgroups.IgnoreNotExists)
```

You could also make your own ErrorHander func and pass to that method if you want it to do other things.

homelessnessbo:
oh,  `control.Stat(cgroups.ErrorHandler(cgroups.IgnoreNotExist))` did the trick! Thx, and sorry for taking your time with my inattention: didnt open `errors.go` file before.

[...]

PR #26, opened 2017-08-24
Extend README with IgnoreNotExists example

onorua:
I've just hit the "bug" https://github.com/containerd/cgroups/issues/5. You had plans to update documentation to prevent confusion, but it seems you had no time to do that. 

This PR should address missing example of ignore missing error handling.


crosbymichael:
LGTM

Thanks!

PR #28, opened 2017-09-05
Use protobufs as Stats return result

crosbymichael:


mlaventure:
- May need a makefile or a script to generate the protos correctly.
- Maybe have it as `v1` ?

metrics.proto
```diff
@@ -0,0 +1,107 @@
...
+	uint64 pg_maj_fault = 10;
+	uint64 inactive_anon = 11;
+	uint64 active_anon = 12;
+	uint64 inactive_file = 13;
+	uint64 active_file = 14;
+	uint64 unevictable = 15;
+	uint64 hierarchical_memory_limit = 16;
+	uint64 hierarchical_swap_limit = 17;
+	uint64 total_cache = 18;
+	uint64 total_rss = 19;
```
stevvooe: Use `gogo.customname` to remove the case flip (RSS -> Rss) above. Let me know if you need an example.

stevvooe:
Presumably, this is using `protobuild` with no configuration?

metrics.proto
crosbymichael: I thought we had support in protobuild for this?  Because we removed all the `customname` tags in the proto files and it just does it for us now

crosbymichael:
@stevvooe so I added customname and protobuild but i'm getting a build error

```
protobuild 
protoc -I.:/home/michael/development/gocode/src:/usr/local/include --include_imports --descriptor_set_out=/tmp/descriptors.pb-973797639 --gogoctrd_out=plugins=grpc,import_path=github.com/containerd/cgroups,Mgogoproto/gogo.proto=github.com/gogo/protobuf/gogoproto,Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/descriptor.proto=github.com/gogo/protobuf/protoc-gen-gogo/descriptor,Mgoogle/protobuf/field_mask.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types:/home/michael/development/gocode/src /home/michael/development/gocode/src/github.com/containerd/cgroups/metrics.proto
gogoproto/gogo.proto: File not found.
github.com/containerd/cgroups/metrics.proto: Import "gogoproto/gogo.proto" was not found or had errors.
```

I'm not sure what is going on.  This pkg does not have a vendor but I have have gogo in my GOPATH.

stevvooe:
@crosbymichael This is a use case that protobuild doesn't yet solve. Right now, we have a `vendored` relative include set but not a `package` relative include set.

https://github.com/stevvooe/protobuild/pull/10 should address this. Give it a try.

crosbymichael:
@stevvooe updated with the latest changes

metrics.proto
```diff
@@ -0,0 +1,109 @@
+syntax = "proto3";
+
+package io.containerd.cgroups.v1;
+
+import "gogoproto/gogo.proto";
+
+message Metrics {
+	map<string, HugetlbStat> hugetlb = 1;
```
stevvooe: Does this map page size to the metrics? Should that just a field?

metrics.proto
```diff
@@ -0,0 +1,109 @@
...
+}
+
+message MemoryEntry {
+	uint64 limit = 1;
+	uint64 usage = 2;
+	uint64 max = 3;
+	uint64 failcnt = 4;
+}
+
+message BlkioStat {
```
stevvooe: Should this be `BlkIO`?

metrics.proto
```diff
@@ -0,0 +1,109 @@
...
+	uint64 max = 2;
+	uint64 failcnt = 3;
+}
+
+message PidsStat {
+	uint64 current = 1;
+	uint64 limit = 2;
+}
+
+message CpuStat {
```
stevvooe: CPUStat

metrics.proto
```diff
@@ -0,0 +1,109 @@
...
+	CpuUsage usage = 1;
```
stevvooe: CPUUsage

metrics.proto
crosbymichael: Yes, page sizes

stevvooe:
LGTM

A few nits.

metrics.proto
crosbymichael: I'm not sure it can be a field because they can change based on a system

stevvooe: I was just thinking a string field `pagesize` instead of a map.

crosbymichael: we can do that

PR #30, opened 2017-10-05
Distinguish not exist error returned by `Path`.

Random-Liu:
@crosbymichael `path` may also return not exist error, e.g. `/proc/PID/cgroup` may not exist.

We should also return `ErrCgroupDeleted` in that case.

Signed-off-by: Lantao Liu <lantaol@google.com>


crosbymichael:
LGTM

PR #34, opened 2017-11-28
Update doc for OOMEventFD

dnephin:
Small change to document this error return value, since it seemed to be relevant in https://github.com/containerd/containerd/pull/1803 it may be relevant to other consumers.


crosbymichael:
LGTM

PR #37, opened 2018-02-22
Fix tests when there's no cgroup hierarchy mounted

elboulangero:
This happens when building the package in a chroot. To be more accurate, when building the package for Debian and using pbuilder.


crosbymichael:
LGTM

PR #45, opened 2018-03-23
Add Update method for the cpuset controller

anastop:
So that the CPUs and memory nodes of a cgroup can be dynamically changed.

### DIFF
--- a/cpuset.go
+++ b/cpuset.go
@@ -82,6 +82,10 @@ func (c *cpusetController) Create(path string, resources *specs.LinuxResources) 
 	return nil
 }
 
+func (c *cpusetController) Update(path string, resources *specs.LinuxResources) error {
+	return c.Create(path, resources)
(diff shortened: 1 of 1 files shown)


crosbymichael:
Have you tried this out to see if it works or if there were any side effects?

anastop:
I have tried this in my case (dynamic process CPU rescheduling) and it works as expected, that is, processes migrate to the new cpuset's CPUs. Unfortunately, my test environment (single-NUMA node) does not allow me to test cpuset.mems. Other than this, are there particular test cases I could run to see possible side effects?

crosbymichael:
LGTM

crosbymichael:
@anastop can you sign your commit with the DCO ?

anastop:
should be ok now

crosbymichael:
LGTM

Issue #49, opened 2018-05-10
Add a new Cgroup interface AddTask()?


jingxiaolu:
Hi Containerd Authors,

In interface Cgroup we only have [`Add(Process)`](https://github.com/containerd/cgroups/blob/07683a668a6c20f509112ea932da7e4ca98c7c3f/control.go#L52), which operates on [`cgroup.procs`](https://github.com/containerd/cgroups/blob/07683a668a6c20f509112ea932da7e4ca98c7c3f/cgroup.go#L130).
Sometimes we want to set tasks (thread ID), but now no interface can be used.

I propose adding a new interface `AddTask()` to work on it, what's your opinions?

I can work on it if you think it is reasonable :-)

jingxiaolu:
@crosbymichael thanks a lot!
PR is ready now at [#50](https://github.com/containerd/cgroups/pull/50), please help to review ~

PR #50, opened 2018-05-11
Add interface AddTask to control groups.

jingxiaolu:
So that we can set tasks when we need.

Signed-off-by: Lu Jingxiao <lujingxiao@huawei.com>

Closes: #49 

Thanks everyone  :smiley:

### DIFF
--- a/control.go
+++ b/control.go
@@ -24,6 +24,7 @@ import (
 
 const (
 	cgroupProcs    = "cgroup.procs"
+	cgroupTasks    = "tasks"
 	defaultDirPerm = 0755
 )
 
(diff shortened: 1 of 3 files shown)


crosbymichael:
LGTM

jingxiaolu:
Thanks a lot @crosbymichael :smiley:  
So one more reviewer is needed now?

control.go
```diff
@@ -48,8 +49,10 @@ type Process struct {
 type Cgroup interface {
 	// New creates a new cgroup under the calling cgroup
 	New(string, *specs.LinuxResources) (Cgroup, error)
-	// Add adds a process to the cgroup
+	// Add adds a process to the cgroup (cgroup.procs)
 	Add(Process) error
+	// Add adds a process to the cgroup (tasks)
```
estesp: The comment should be `AddTask adds a process..`; we don't have the linter running here, but that would be required for the comment to match the function name

jingxiaolu: Thanks @estesp :+1: , code updated :smile: , please help to check again~

PR #51, opened 2018-08-13
Fix empty device type

crosbymichael:
Empty is a common convention for all in oci


dmcgowan:
LGTM

PR #53, opened 2018-09-05
systemd-239+ no longer allows delegate slice

baude:
Systemd no longer allows the setting of the Delegate property for
cgroups.  This patch is modeled after a runc commit 0e16bd9b53eb3c57ea6fe59fc6d9385c2edb9fd9
which corrects a similar situation for them.

Signed-off-by: baude <bbaude@redhat.com>

### DIFF
--- a/systemd.go
+++ b/systemd.go
@@ -32,6 +32,11 @@ const (
 	defaultSlice      = "system.slice"
 )
 
+var (
+	canDelegate bool
+	once sync.Once
+)
+
 func Systemd() ([]Subsystem, error) {
 	root, err := v1MountPoint()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mrunalp:
@crosbymichael ptal.

systemd.go
```diff
@@ -80,15 +80,36 @@ func (s *SystemdController) Create(path string, resources *specs.LinuxResources)
 	}
 	defer conn.Close()
 	slice, name := splitName(path)
+	// We need to see if systemd can handle the delegate property
+	// Systemd will return an error if it cannot handle delegate regardless
+	// of its bool setting.
+	canDelegate := true
```
crosbymichael: can we move this to a `sync.Once` ?

baude:
@crosbymichael is this what you had in mind ?

systemd.go
```diff
@@ -78,17 +82,42 @@ func (s *SystemdController) Create(path string, resources *specs.LinuxResources)
 	if err != nil {
 		return err
 	}
+	var once sync.Once
```
crosbymichael: This should be declared at the package level but otherwise, looks good

baude: Ok, fixed

dmcgowan:
LGTM

crosbymichael:
LGTM

PR #63, opened 2018-10-25
Makes Load function more lenient on subsystems' checking

ChrsMark:
Currently when a cgroup path is not found under a subsystem's directory
Load function will fail completely. This can be blocking when a subsystem
is taking time to update its cgroups. This commit makes Load to by-pass
any not-found subsystem returning a Cgroup with a list of the
only-found subsystems.


crosbymichael:
Thanks, i'll try to get this reviewed today or tomorrow

cgroup.go
```diff
@@ -52,8 +52,9 @@ func Load(hierarchy Hierarchy, path Path) (Cgroup, error) {
 	if err != nil {
 		return nil, err
 	}
-	// check the the subsystems still exist
-	for _, s := range pathers(subsystems) {
```
crosbymichael: could we keep the `for range` the same and create a temporary var for valid subsystems?  Maybe something like `activeSubsystems` and you append active ones instead of removing the ones that are not found from the current list? I think it would make the code simpler and explicit.

What do you think?

crosbymichael:
Also, after you make the updates, do you think you could work up a test in a separate commit to verify that we handle omitting a subsystem?

cgroup.go
ChrsMark: Sure! I had the same thought but chose to go with the more "economical" way by keeping the same array.

crosbymichael:
@ChrsMark Thanks!  The code is looking great now!

crosbymichael:
@ChrsMark when you update and add the test, can you rebase on master so that the CI runs successfully?

crosbymichael:
LGTM

Thanks @ChrsMark and @MichaelKatsoulis !

PR #67, opened 2018-11-21
Add functionality for retrieving all tasks of a cgroup

gpanouts:
So that we can get all tasks of a certain cgroup when we need to. Until now it has been possible to get only processes of that cgoup. 

@crosbymichael 
@estesp


crosbymichael:
LGTM

Thanks @gpanouts , code looks great!

PR #72, opened 2018-12-09
Add tests for pids

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

### DIFF
--- a/pids_test.go
+++ b/pids_test.go
@@ -1,0 +1,172 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cgroups
+
+import (
+	"encoding/hex"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


pids_test.go
```diff
@@ -0,0 +1,169 @@
...
+		bytes,
+		defaultFilePerm,
+	); err != nil {
+		t.Error(err)
+		return
+	}
+	metrics := Metrics{}
+	err = pids.Stat("test", &metrics)
+	assert.NotNil(t, err)
+}
```
crosbymichael: Can you `gofmt` this code?

gliptak: updated

pids_test.go
```diff
@@ -0,0 +1,169 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cgroups
+
+import (
+	"encoding/hex"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/stretchr/testify/assert"
```
crosbymichael: I don't think we use this package in containerd and definitely not this one.  Can we remove this?

You also need to use `goimports` to format the imports in this PR

pids_test.go
```diff
@@ -0,0 +1,169 @@
...
+	if pids == nil {
+		t.Error("pids is nil")
+		return
+	}
+	resources := specs.LinuxResources{}
+	resources.Pids = &specs.LinuxPids{}
+	resources.Pids.Limit = int64(10)
+	err = pids.Create("test", &resources)
+	if err != nil {
+		t.Error(err)
```
crosbymichael: You can use `t.Fatal` here so you don't have to have a `return` in every block

gliptak: updated

pids_test.go
gliptak: updated

gliptak:
@crosbymichael  are there other updates you would like to see?

crosbymichael:
LGTM

Thanks for all the updates to this, the code looks really good!

PR #77, opened 2019-02-14
Check for non-active/supported cgroups

### LINKED ISSUE #76: unable to find * in controller set: unknown
There are problems with this library handling missing cgroups or the way they are mounted. The unhandled configuration seems to become common with new systemd.

These issues list the output of mount and proc in the affected systems:

https://github.com/docker/for-linux/issues/545
https://github.com/docker/for-linux/issues/552

@crosbymichael

crosbymichael:
Fixes #76

I still need to get a system setup to test this fix.

### DIFF
--- a/cgroup.go
+++ b/cgroup.go
@@ -30,7 +30,13 @@ import (
 )
 
 // New returns a new control via the cgroup cgroups interface
-func New(hierarchy Hierarchy, path Path, resources *specs.LinuxResources) (Cgroup, error) {
+func New(hierarchy Hierarchy, path Path, resources *specs.LinuxResources, opts ...InitOpts) (Cgroup, error) {
+	config := newInitConfig()
+	for _, o := range opts {
+		if err := o(config); err != nil {
+			return nil, err
+		}
+	}
 	subsystems, err := hierarchy()
 	if err != nil {
 		return nil, err
@@ -40,6 +46,13 @@ func New(hierarchy Hierarchy, path Path, resources *specs.LinuxResources) (Cgrou
 		// check if subsystem exists
 		if err := initializeSubsystem(s, path, resources); err != nil {
 			if err == ErrControllerNotActive {
+				if config.InitCheck != nil {
+					if skerr := config.InitCheck(s, path, err); skerr != nil {
+						if skerr != ErrIgnoreSubsystem {
+							return nil, skerr
+						}
+					}
+				}
 				continue
 			}
 			return nil, err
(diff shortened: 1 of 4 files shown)


crosbymichael:
I was able to verify this on a system that Tonis setup for me.  I'm happy with these changes and this is good to review.

```bash
>  docker run alpine true
docker: Error response from daemon: controller is not supported: unknown.
ERRO[0000] error waiting for container: context canceled

>  cp * /usr/bin/ #patched containerd
> docker run alpine true
> docker run alpine ls
bin
dev
etc
home
lib
media
mnt
opt
proc
root
run
sbin
srv
sys
tmp
usr
var
```

cgroup.go
```diff
@@ -60,6 +77,16 @@ func Load(hierarchy Hierarchy, path Path) (Cgroup, error) {
 			if os.IsNotExist(errors.Cause(err)) {
 				return nil, ErrCgroupDeleted
 			}
+			if err == ErrControllerNotActive {
```
Ace-Tang: Do we need to change the location for `if os.IsNotExist(errors.Cause(err))` and  `if err == ErrControllerNotActive`, like
```
			if err == ErrControllerNotActive {
				for _, o := range skip {
					if skerr := o(s, path, err); skerr != nil {
						if skerr != ErrSkipSubsystem {
							return nil, skerr
						}
					}
				}
				continue
			}
                        if os.IsNotExist(errors.Cause(err)) {
				return nil, ErrCgroupDeleted
			}
```
first check controller is support, then check the file is exist ?

Ace-Tang:
Overall looks really good to me, from reading the patch, I think it can solve my problem (I mount some controller like pid, cpu, memory through cgroup2)

cgroup.go
crosbymichael: I think this is fine to keep the existing logics for Load, the isnotexist is for when the cgroup directory structure has been removed, not an active controller.  I think this is fine for the Load case to get the ErrCgroupDeleted error which is handled

cgroup.go
```diff
@@ -30,24 +30,41 @@ import (
...
+				continue
+			}
 			return nil, err
 		}
+		active = append(active, s)
 	}
 	return &cgroup{
 		path:       path,
-		subsystems: subsystems,
+		subsystems: active,
```
mikebrow: how about adding a list for skipped .. or inactive?

cgroup.go
```diff
@@ -30,24 +30,41 @@ import (
...
 	}, nil
 }
 
 // Load will load an existing cgroup and allow it to be controlled
-func Load(hierarchy Hierarchy, path Path) (Cgroup, error) {
+func Load(hierarchy Hierarchy, path Path, skip ...SkipOpts) (Cgroup, error) {
```
mikebrow: ditto for update to function description

paths.go
```diff
@@ -57,6 +57,9 @@ func PidPath(pid int) Path {
 	return existingPath(paths, "")
 }
 
+// ErrControllerNotActive is returned when a controller is not supported or enalbed
```
mlaventure: `s/enalbed/enabled/`

cgroup.go
crosbymichael: not sure we need it

PR #79, opened 2019-03-28
Return ErrCgroupDeleted when no subsystems

crosbymichael:
Return the error when 0 subsystems are active signaling that the cgroup
has been deleted.

Fix for: https://github.com/containerd/containerd/issues/3133


dmcgowan:
LGTM

Issue #80, opened 2019-04-12
Add network metrics


ZeroMagic:
When we use containerd to run our applications, we can not collect the network metrics about containers. Can we add the network metrics into metrics.proto? 

```go
type NetworkInterface struct {
	// Name is the name of the network interface.
	Name string

	RxBytes   uint64
	RxPackets uint64
	RxErrors  uint64
	RxDropped uint64
	TxBytes   uint64
	TxPackets uint64
	TxErrors  uint64
	TxDropped uint64
}
```

ping @crosbymichael

crosbymichael:
Where are these metrics going to come from?

ZeroMagic:
> Where are these metrics going to come from?

@crosbymichael 
Currently KataContainers collect the metrics from [libcontainer Stats() API](https://github.com/opencontainers/runc/blob/master/libcontainer/container_linux.go#L175). 

The libcontainer Stats() API has collected the network metrics. However KataContainers do not transfer the network metrics now. I will propose a similar issue in KataContainers community.

Then the network metrics will come from [shimV2 Stats() API](https://github.com/kata-containers/runtime/blob/master/containerd-shim-v2/service.go#L776).

crosbymichael:
The reason why I kept network stats out of this pkg is because a lot of the network stats come from `/sys` and not cgroups

ZeroMagic:
I got it. Where do you think I should add network stats?  
Sincerely looking forward to some guidance.

thaJeztah:
@ZeroMagic with https://github.com/containerd/cgroups/pull/81 merged; is this one ready too close?

PR #81, opened 2019-04-17
Add network stats

ZeroMagic:
Ref: #80 
Actually, containerd has no concept of networks. Then containerd only needs to transfer the network stats to CRI or other receivers. So just add the netstat stuff to the cgroups repo for now.

Signed-off-by: ZeroMagic <anthonyliu@zju.edu.cn>


metrics.proto
```diff
@@ -121,3 +122,15 @@ message RdmaEntry {
 	uint32 hca_handles = 2;
 	uint32 hca_objects = 3;
 }
+
+message NetworkStat {
+	string name = 1;
+	uint64 RxBytes = 2;
```
crosbymichael: these should be lowercase:

`rx_bytes` `tx_bytes`

ZeroMagic: done.

crosbymichael:
LGTM

ZeroMagic:
@fuweid Can you help with the second round of review?

fuweid:
maybe we can put the proto version in READMD.md so that it is easy to make the right pb.go file. :p

ZeroMagic:
The version of gogo/protobuf I use is v0.5. Its commit id is 342cbe0a04158f6dcb03ca0079991a51a4248c02.

stormgbs:
LGTM

fuweid:
The pb file is correct and it is the same with https://github.com/containerd/containerd/blob/master/script/setup/install-protobuf. 

ping @AkihiroSuda could you help us to review it again? thanks

crosbymichael:
LGTM

PR #82, opened 2019-05-15
Add go module support

### LINKED ISSUE #46: tagged release versions
Would it be possible to have tagged versioned releases for this repo? Is there a reason why we're explicitly choosing to not have versioned releases? LMK if it's just a matter of doing the work, I could make those changes.

I think it would improve ux for at least those projects using dep to manage containerd/cgroups as a dependency.

saschagrunert:
This commit adds a `go.mod` with the latest dependencies. It would be awesome if we could tag a new semver release version after this PR got merged. :innocent: 

Closes #46


fiws:
..Why is this still not merged?

crosbymichael:
LGTM

PR #88, opened 2019-06-13
cgroups: fix MoveTo function fail problem

woshijpf:
reason: there maybe some multi processes in the cgroup.procs file,if
some processes exit before call destination.Add(p), it will get the error
"no such process" which means process has exit and we just ignore this
error and continue to move left processes in cgroup.procs file.


crosbymichael:
Can you sign your commit with the DCO?  `git commit -s`?

woshijpf:
@crosbymichael Thanks for reminding, I have pushed a new commit with signed-off info.

crosbymichael:
LGTM

PR #96, opened 2019-09-17
Move metrics proto package to stats/v1

jterry75:
In order for projects to link the metrics.proto file with go modules it
is required to have a go import. This creates a problem because the
cgroups package is not platform independent although its types are.
Moving the types to a sperate package allows consumption of the types
regardless of platform.

Signed-off-by: Justin Terry (VM) <juterry@microsoft.com>


crosbymichael:
change looks good but tests need updated

jterry75:
Oops. I thought for sure I ran these.

crosbymichael:
LGTM

PR #97, opened 2019-09-21
Allow overriding the default /proc folder in blkioController

mariomac:
A host may share its `/proc` FS folder with a privileged container, mounted in an
alternative folder (e.g. `/host/proc`), as some users prefer to run their
monitoring software inside a container rather than a host process.

This patch allows overriding the default `/proc` folder to allow the blkio
Controller working inside containerized monitoring software.

The `/proc` root can be overriden via an optional function (`ProcRoot`) to
avoid introducing a new `NewBlkio` constructor or introducing breaking changes
in the current API.

Then, the `NewBlkio` constructor can be invoked as usual, or in the following
form, to override the /proc path:

```go
ctrl := NewBlkio("/sys/fs/cgroup", ProcRoot("/host/proc"))
```

### DIFF
--- a/blkio.go
+++ b/blkio.go
@@ -30,14 +30,29 @@ import (
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
 
-func NewBlkio(root string) *blkioController {
-	return &blkioController{
-		root: filepath.Join(root, string(Blkio)),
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

PR #98, opened 2019-09-23
Allows ignoring memory modules

mariomac:
Some hosts may have disabled some memory modules (e.g. `memsw` metrics
are not available if the Kernel is built without the
`CONFIG_MEMCG_SWAP_ENABLED` variable. This made the memory controller
`Stat` function to not work properly, because it returns error
when a single memory entry is not available.

Even if you run `cgroups.Stat(cgroups.IgnoreNotExist)` to ignore memory
Stat errors, you won't get most of the memory metrics.

This patch adds a configuration function parameter to the `NewMemory`
constructor. This way, we don't introduce breaking changes in the
current API. You can invoke the `NewMemory` constructor as usual or
with a new optional configuration option:

```go
mem := cgroups.NewMemory(root, cgroups.IgnoreModules("memsw"))
```

If you want to let the constructor to automatically ignore `memsw` module
only if it is not available:

```go
mem := cgroups.NewMemory(root, cgroups.OptionalSwap())
```

Signed-off-by: Mario Macias <mmacias@newrelic.com>

### DIFF
--- a/memory.go
+++ b/memory.go
@@ -32,14 +32,32 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func NewMemory(root string) *memoryController {
-	return &memoryController{
-		root: filepath.Join(root, string(Memory)),
+// NewMemory returns a Memory controller given the root folder of cgroups.
+// It may optionally accept other configuration options, such as IgnoreModules(...)
+func NewMemory(root string, options ...func(*memoryController)) *memoryController {
+	mc := &memoryController{
+		root:    filepath.Join(root, string(Memory)),
+		ignored: map[string]struct{}{},
+	}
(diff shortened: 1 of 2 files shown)


crosbymichael:
Could you make the Ignore more dynamic by checking for the memsw file as an Opt?  You can keep the current IgnoreModule the same but add another opt that can be used in containerd that will add the ignore only if a stat() on memsw fails?

mariomac:
Thanks for the quick feedback @crosbymichael . I submitted a new patch with the `OptionalSwap` constructor option.

Feel free to suggest any other name.

crosbymichael:
LGTM

Thanks @mariomac !!

PR #100, opened 2019-10-09
stats/v1: export per-cgroup stats

fuweid:
The cgroupstats provides information from taskstats which help us to
know how `busy` tasks in one cgroup are right now. With Metric API in
shim, it can enhance monitor capability.

For more information:

* https://github.com/torvalds/linux/blob/master/include/uapi/linux/cgroupstats.h
* https://www.kernel.org/doc/Documentation/accounting/cgroupstats.txt
* https://www.kernel.org/doc/Documentation/accounting/taskstats.txt

Signed-off-by: Wei Fu <fuweid89@gmail.com>


crosbymichael:
So these stats are provided via a netlink socket?

fuweid:
> So these stats are provided via a netlink socket?

Yes. It seems efficient I think.

fuweid:
> Where is the implementation going to come from? the Shims?

Yes. It is for kata container. We put the collector in the agent. It will export the data through Mertic Shim API.

Ace-Tang:
We decide to use metric-server get load information instead of using cadvisor. So add a interface in containerd/cgroup libbrary.

crosbymichael:
LGTM

PR #105, opened 2019-11-04
Update doc add static path prefix

xianlubird:
I use the function `cgroups.Load(cgroups.V1, cgroups.StaticPath(location))`, find that it always report that `cgroup deleted`.  I find that my location is ` /sys/fs/cgroup/cpu/kubepods.slice/kubepods-besteffort.slice/`, it start with `/sys/fs`, through the code I find that, this library will add `/sys/fs/cgroup` prefix auto. So I add the pr to make it more clearly

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,10 @@ err := control.MoveTo(destination)
 subCgroup, err := control.New("child", resources)
 ```
 
+### Attention
+
+All static path should not include `/sys/fs/cgroup/` prefix, it should start with your own cgroups name
+
 ## Project details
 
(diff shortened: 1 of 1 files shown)


xianlubird:
@AkihiroSuda agree, I will add one to godoc, but in readme maybe also need one .  WDYT?

AkihiroSuda:
SGTM

AkihiroSuda:
And please sign the commit

xianlubird:
@AkihiroSuda Already sign

AkihiroSuda:
Please sign with your full name, then LGTM

xianlubird:
Done

crosbymichael:
LGTM

Issue #104, opened 2019-10-31
cgroup2: check list for v2 controller implementations


AkihiroSuda:
## Urgent (required to get CRI and Moby work)
- [x] pids: #103
- [x] cpu: #107
- [x] memory (`memory.stats`): #108 @Zyqsempai
- [x] `memory.current`, `memory.swap.current`, and other memory stat files  https://www.kernel.org/doc/Documentation/cgroup-v2.txt : #117 
- [x] OOM event FD: #115
- [x] io: #112
- [x] devices (BPF): #116
- [x] `ToggleControllers()` (issue: https://github.com/containerd/cgroups/issues/121): #124

Metrics functions are urgent.
`Update()` functions are not necessary for CRI and Moby.

## Non-urgent
- [x] `NewSystemd()` and `LoadSystemd()`: #125 
- [x] freezer: #103
- [ ] perf_event
- [x] cpuset: #111
- [x] rdma: #119 
- [x] [hugetlb](https://www.spinics.net/lists/cgroups/msg24003.html) (to appear in kernel 5.XX)

Zyqsempai:
@AkihiroSuda Can I help you somehow on this one?

AkihiroSuda:
thanks, which do you plan to work on?

Zyqsempai:
@AkihiroSuda I will take memory part if you don't mind.

AkihiroSuda:
thanks, please open PR when #103 gets merged

AkihiroSuda:
#103 merged

Zyqsempai:
@crosbymichael So what do you have in plans? Should I open PR with the next part(memory) ? add then we will add a manager above it?

crosbymichael:
@Zyqsempai sure, go ahead and open more if you still have more coming.

Thanks!

Zyqsempai:
Guys, I will also take the rest if you don't mind!

Zyqsempai:
So guys, let's summarize what we still have to do here?

crosbymichael:
We need OOM notifications from the memory controller as well as the bpf devices code added to this repo

AkihiroSuda:
Do we need devices controller for CRI and Moby?

Zyqsempai:
@crosbymichael @AkihiroSuda I will take OOM notifications and bpf devices.

crosbymichael:
@AkihiroSuda i don't want to put out any cgroups code without device support as it's very important to security

[...]

PR #109, opened 2019-11-05
Simpler v2 cgroup interface

crosbymichael:
The resource type is changed in this but I would rather have a v2 native
Resource type and do a mapping from an OCI spec's resource to the v2.Resource
type.

ref #104 


crosbymichael:
@AkihiroSuda and @Zyqsempai please take a look at this new approach.  it should make things much easier to manage and extend.

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		r.Memory,
+		r.Pids,
+	}
+	for _, v := range values {
+		if v == nil {
+			continue
+		}
+		o = append(o, v.Values()...)
+	}
+	return o
```
Zyqsempai: Can you make the name of this variable more explicit?

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return ioutil.WriteFile(
+		filepath.Join(path, c.key),
+		data,
+		perm,
+	)
+}
+
+func writeCGStr(path string, strs []CGStr) error {
+	for _, o := range strs {
```
Zyqsempai: Same here, actually I can't get what `o` stands for;)

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		if err := o.write(path, defaultFilePerm); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func NewController(mountpoint string, g string, resources *Resources) (*Controller, error) {
```
Zyqsempai: Same, since `g` not any more of the GroupPath type, it's not so obvious what it stands for.

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, v := range values {
+		if v == nil {
+			continue
+		}
+		o = append(o, v.Values()...)
+	}
+	return o
+}
+
+type CGStr struct {
```
Zyqsempai: Also, the naming of this struct raises some questions.

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		if current == state {
+			return nil
+		}
+		time.Sleep(1 * time.Millisecond)
+	}
+}
+
+func (c *Controller) Stat(g string) (*statsv2.Metrics, error) {
+	var stats statsv2.Metrics
+	return &stats, nil
```
Zyqsempai: Sorry, I didn't get it, how do you plan to collect stats for different controllers?

v2/controller.go
fuweid: it is ok to use short name here. `o` sounds like output.:P

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+
+	path := filepath.Join(mountpoint, string(g))
+	if err := os.MkdirAll(path, defaultDirPerm); err != nil {
+		return nil, err
+	}
+	if err := writeCGStr(path, resources.Values()); err != nil {
+		// clean up cgroup dir on failure
+		os.Remove(path)
```
fuweid: should be remove folder just in case it is not empty dir.

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+	if strings.HasPrefix(name, "/") {
+		return nil, errors.New("name must be relative")
+	}
+	path := filepath.Join(c.path, name)
+	if err := os.MkdirAll(path, defaultDirPerm); err != nil {
+		return nil, err
+	}
+	if err := writeCGStr(path, resources.Values()); err != nil {
+		// clean up cgroup dir on failure
+		os.Remove(path)
```
fuweid: should be remove folder by `os.RemoveAll` just in case it is not empty dir.

v2/controller.go
fuweid: cgroupv2 will list enabled controllers in `cgroup.controllers` file. I think it is easy to collect the data for the controllers.

v2/utils.go
```diff
@@ -167,7 +148,7 @@ func parseCgroupFromReader(r io.Reader) (GroupPath, error) {
 		}
 		// text is like "0::/user.slice/user-1001.slice/session-1.scope"
 		if parts[0] == "0" && parts[1] == "" {
-			return GroupPath(parts[2]), nil
```
fuweid: still feel that it is not more explicit. the pattern is `0::PATH` and not need to split it into 3 parts😂

v2/controller.go
AkihiroSuda: `Value`?

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return o
+}
+
+type CGStr struct {
+	key   string
+	value interface{}
+}
+
+func (c *CGStr) write(path string, perm os.FileMode) error {
```
AkihiroSuda: please add a code comment to clarify `path` is file path (including `/sys/fs/cgroup` prefix), not a group path?

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+	path := filepath.Join(mountpoint, string(g))
+	return &Controller{
+		unifiedMountpoint: mountpoint,
+		path:              path,
+	}, nil
+}
+
+type Controller struct {
```
AkihiroSuda: Manager?

v2/paths.go
```diff
@@ -22,33 +22,26 @@ import (
...
-func PidGroupPath(pid int) (GroupPath, error) {
+func PidGroupPath(pid int) (string, error) {
 	p := fmt.Sprintf("/proc/%d/cgroup", pid)
 	return parseCgroupFile(p)
 }
 
 // VerifyGroupPath verifies the format of g.
 // VerifyGroupPath doesn't verify whether g actually exists on the system.
-func VerifyGroupPath(g GroupPath) error {
+func VerifyGroupPath(g string) error {
```
AkihiroSuda: Please either move `GroupPath` godoc here, or restore `GroupPath` type?

```
// GroupPath is a string that appears as the third field in /proc/PID/cgroup.
// e.g. "/user.slice/user-1001.slice/session-1.scope"
//
// GroupPath must not contain "/sys/fs/cgroup" prefix.
// GroupPath must be a absolute path starts with "/".
```

v2/cpu.go
```diff
@@ -16,108 +16,23 @@
...
+	if r.Weight != nil {
+		o = append(o, CGStr{
+			key:   "cpu.weight",
+			value: *r.Weight,
+		})
 	}
-	if cpuShares := resources.CPU.Shares; cpuShares != nil {
-		// Converting cgroups configuration from v1 to v2
-		// more here https://github.com/containers/crun/blob/master/crun.1.md#cgroup-v2
-		convertedWeight := (1 + ((*cpuShares-2)*9999)/262142)
```
AkihiroSuda: where is the new code for this?

v2/controller.go
```diff
@@ -0,0 +1,212 @@
...
+func (c *Controller) AddProc(pid uint64) error {
+	v := CGStr{
+		key:   cgroupProcs,
+		value: pid,
+	}
+	return writeCGStr(c.path, []CGStr{v})
+}
+
+func (c *Controller) Delete() error {
+	return remove(c.path)
```
AkihiroSuda: can be just `os.Remove`?

AkihiroSuda:
`cmd/cgroups-playground` needs to be updated

v2/controller.go
crosbymichael: In cgroups, it's usually just a `rm`

v2/controller.go
crosbymichael: As for the implementation, i'm still working on a simpler way to the files and parse them easily without a lot of switches.  For now, we can just omit stats right now.

v2/cpu.go
crosbymichael: i'm skipping this for now to make sure we have a simple API first, then we can port the spec -> resource type code over along with stats.  The old API is too complicated and has too much code.

v2/controller.go
crosbymichael: I think we still need the remove all logic but i'm not sure if this is more robust in v2

v2/manager.go
```diff
@@ -0,0 +1,211 @@
...
+
+type Manager struct {
+	unifiedMountpoint string
+	path              string
+}
+
+func (c *Manager) NewChild(name string, resources *Resources) (*Manager, error) {
+	if strings.HasPrefix(name, "/") {
+		return nil, errors.New("name must be relative")
+	}
```
AkihiroSuda: write `cgroup.subtree_control`?

v2/paths.go
```diff
@@ -22,33 +22,26 @@ import (
...
 	s := string(g)
```
AkihiroSuda: can be removed

v2/paths.go
```diff
@@ -22,33 +22,26 @@ import (
...
 // PidGroupPath will return the correct cgroup paths for an existing process running inside a cgroup
 // This is commonly used for the Load function to restore an existing container
-func PidGroupPath(pid int) (GroupPath, error) {
+func PidGroupPath(pid int) (string, error) {
 	p := fmt.Sprintf("/proc/%d/cgroup", pid)
 	return parseCgroupFile(p)
 }
 
 // VerifyGroupPath verifies the format of g.
 // VerifyGroupPath doesn't verify whether g actually exists on the system.
```
AkihiroSuda: I still think we need godoc to clarify what is the expected format.
Probably we should add an example string like `"/user.slice/user-1001.slice/session-1.scope"`

v2/memory.go
```diff
@@ -132,11 +60,11 @@ func (m *memoryController) Stat(g GroupPath, stats *statsv2.Metrics) error {
 		if err := sc.Err(); err != nil {
 			return err
 		}
-		key, v, err := parseKV(sc.Text())
+		filename, v, err := parseKV(sc.Text())
```
Zyqsempai: It's not so clear, why `parseKV`(parse Key and value) function returns filename. I think `key` was way more clear.

crosbymichael:
Ok, pushed a big update. This has stats implemented along with renaming the playground tool to `cgctl` with a couple methods to play with this.  

```bash
NAME:
   cgctl - cgroup v2 management tool

USAGE:
   cgctl [global options] command [command options] [arguments...]

VERSION:
   1

COMMANDS:
   new      create a new cgroup
   del      delete a cgroup
   list     list processes in a cgroup
   stat     stat a cgroup
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --debug             enable debug output in the logs
   --mountpoint value  cgroup mountpoint (default: "/sys/fs/cgroup")
   --help, -h          show help
   --version, -v       print the version
```

```bash
> sudo ./cgctl new --enable test
> sudo ./cgctl list test
> sudo ./cgctl stat test
> sudo ./cgctl del test
```

v2/paths.go
AkihiroSuda: I meant type casting can be removed. The function should be kept?

cmd/cgctl/main.go
```diff
@@ -0,0 +1,141 @@
...
+		c, err := v2.NewManager(clix.GlobalString("mountpoint"), path, nil)
+		if err != nil {
+			return err
+		}
+		if clix.Bool("enable") {
+			controllers, err := c.ListControllers()
+			if err != nil {
+				return err
+			}
+			if err := c.ToggleControllers(controllers, v2.Enable); err != nil {
```
AkihiroSuda: This needs to be done for the parent group?

v2/paths.go
```diff
@@ -19,42 +19,21 @@ package v2
...
 // This is commonly used for the Load function to restore an existing container
-func PidGroupPath(pid int) (GroupPath, error) {
+func PidGroupPath(pid int) (string, error) {
 	p := fmt.Sprintf("/proc/%d/cgroup", pid)
 	return parseCgroupFile(p)
 }
-
-// VerifyGroupPath verifies the format of g.
-// VerifyGroupPath doesn't verify whether g actually exists on the system.
-func VerifyGroupPath(g GroupPath) error {
```
AkihiroSuda: this function can be kept

cmd/cgctl/main.go
crosbymichael: you don't need to enable this at all if the system is setup to enable them

AkihiroSuda: When you have `cpu` in the top group `/sys/fs/cgroup`, `cpu` is enabled for `/sys/fs/cgroup/foo`, but not enabled for `/sys/fs/cgroup/foo/bar`.

```console
root@suda-ws01:/sys/fs/cgroup# cat cgroup.controllers 
cpuset cpu io memory pids rdma
root@suda-ws01:/sys/fs/cgroup# mkdir foo
root@suda-ws01:/sys/fs/cgroup# cat foo/cgroup.controllers 
cpu io memory pids
root@suda-ws01:/sys/fs/cgroup# mkdir foo/bar
root@suda-ws01:/sys/fs/cgroup# cat foo/bar/cgroup.controllers 
root@suda-ws01:/sys/fs/cgroup#
```

v2/manager.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func (c *Manager) Stat() (map[string]uint64, error) {
+	controllers, err := c.ListControllers()
+	if err != nil {
+		return nil, err
+	}
+	out := make(map[string]uint64)
+	for _, controller := range controllers {
+		filename := fmt.Sprintf("%s.stat", controller)
```
AkihiroSuda: how will it work for other files such as `pids.current`?

AkihiroSuda: Probably we should either add file name to key or return `map[string]map[string]interface{}`

cmd/cgctl/main.go
crosbymichael: yes, its up to the creator of the cgroup to enable these.  I don't think the sub cgroup should modify the parent settings

v2/manager.go
crosbymichael: i'ts still a uint

crosbymichael:
We don't have to fix everything in this PR, it's meant to show that we can simplify things for v2 and to get this in so we can carry on the work

AkihiroSuda:
follow-up: https://github.com/containerd/cgroups/pull/110

PR #110, opened 2019-11-07
 v2: restore VerifyGroupPath;  v2: fix TestParseCgroupFromReader

AkihiroSuda:
restore `VerifyGroupPath` removed in https://github.com/containerd/cgroups/pull/109#discussion_r343218221

Also fixes failing test
```
    --- FAIL: TestParseCgroupFromReader (0.00s)
        utils_test.go:34: expected
    "/user.slice/user-1001.slice/session-1.scope", got ""
        utils_test.go:37: invalid cgroup entry: "2:cpuset:/foo"
```


v2/paths.go
```diff
@@ -37,3 +38,23 @@ func PidGroupPath(pid int) (string, error) {
...
+		return ErrInvalidGroupPath
+	}
+	if filepath.Clean(g) != g {
+		return ErrInvalidGroupPath
+	}
+	if strings.HasPrefix(g, "/sys/fs/cgroup") {
+		return ErrInvalidGroupPath
+	}
+	return nil
+}
```
Zyqsempai: Since we return here same error for all cases, don't you want to use something like:
```
func VerifyGroupPath(g string) error {
	if !strings.HasPrefix(g, "/") ||
	   filepath.Clean(g) != g ||
           strings.HasPrefix(g, "/sys/fs/cgroup") {
		return ErrInvalidGroupPath
	}

	return nil
}
```

AkihiroSuda: I feel it is less readable

crosbymichael:
LGTM

PR #112, opened 2019-11-07
Add v2 io support

crosbymichael:

### DIFF
--- a/v2/manager.go
+++ b/v2/manager.go
@@ -20,12 +20,14 @@ import (
 	"bufio"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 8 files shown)


v2/manager.go
```diff
@@ -245,12 +251,17 @@ func (c *Manager) Procs(recursive bool) ([]uint64, error) {
 	return processes, err
 }
 
-func (c *Manager) Stat() (map[string]uint64, error) {
+var singleValueFiles = []string{
+	"pids.current",
+	"pids.max",
```
AkihiroSuda: can be another PR, but we need `memory.current` and `memory.swap.current`?
https://facebookmicrosites.github.io/cgroup2/docs/memory-controller.html

crosbymichael:
@Zyqsempai this has the code for how we will handle stats

PR #114, opened 2019-11-08
Added additional param to memory and convert value to cpu

Zyqsempai:
Added `memory.kmem.limit_in_bytes` resource to memory and `cpu.weight` conversion for CPU.

### DIFF
--- a/v2/cpu.go
+++ b/v2/cpu.go
@@ -25,9 +25,12 @@ type CPU struct {
 
 func (r *CPU) Values() (o []Value) {
 	if r.Weight != nil {
+		// Converting cgroups configuration from v1 to v2
+		// more here https://github.com/containers/crun/blob/master/crun.1.md#cgroup-v2
+		convertedWeight := (1 + ((*r.Weight-2)*9999)/262142)
 		o = append(o, Value{
 			filename: "cpu.weight",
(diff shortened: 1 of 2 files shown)


Zyqsempai:
@crosbymichael Please take a look, does is how you plan it to do?
Also still open question for me with `Stats` how should we collect them for each particular controller?

Zyqsempai:
@AkihiroSuda It's contra intuitive then, coz I expect to find specific controller logic in the controller code. Where should it be placed then?

v2/cpu.go
```diff
@@ -25,9 +25,12 @@ type CPU struct {
 
 func (r *CPU) Values() (o []Value) {
 	if r.Weight != nil {
+		// Converting cgroups configuration from v1 to v2
+		// more here https://github.com/containers/crun/blob/master/crun.1.md#cgroup-v2
+		convertedWeight := (1 + ((*r.Weight-2)*9999)/262142)
```
crosbymichael: I already did this in another PR, in the utils.go file there is a ToResources that has all the mappings from an OCI spec into the Resource type

Zyqsempai: Oh, got it. Then we can close this one.

PR #115, opened 2019-11-12
Added OOM notification for Memory controller

### LINKED ISSUE #104: cgroup2: check list for v2 controller implementations

Zyqsempai:
Partially Fixes #104 

@crosbymichael @AkihiroSuda 
So this Pr is on of the ideas I have on how we can implement `OOM notifications` in `cGroupv2`.
So, I see here at least two directions, we can implement `Inotify fd` check right here and in case if we receive `inotify` event, create new separate file for `oom's` only, and return that file `fd`.
On other hand we can delegate it to next layers(containerd etc.)
What do you think of this approach?

### DIFF
--- a/v2/manager.go
+++ b/v2/manager.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"math"
 	"os"
 	"path/filepath"
@@ -39,9 +40,13 @@ const (
 	controllersFile = "cgroup.controllers"
 )
 
+var oomEventCount = 0
+
 type cgValuer interface {
 	Values() []Value
 }
+
+type Event struct{}
 
 // Resources for a cgroups v2 unified hierarchy
 type Resources struct {
@@ -415,8 +420,8 @@ func (c *Manager) freeze(path string, state State) error {
 	}
 }
 
-func (c *Manager) OOMEventFD(rootPath string) (uintptr, error) {
-	fpath := filepath.Join(rootPath, "memory.events")
+func (c *Manager) OOMEventFD() (uintptr, error) {
+	fpath := filepath.Join(c.path, "memory.events")
 	fd, err := syscall.InotifyInit()
 	if err != nil {
 		return 0, fmt.Errorf("Failed to create inotify fd")
@@ -429,4 +434,35 @@ func (c *Manager) OOMEventFD(rootPath string) (uintptr, error) {
 	defer syscall.InotifyRmWatch(fd, uint32(wd))
 
 	return uintptr(fd), nil
+}
+
+func (c *Manager) EventChan() (<-chan Event, error) {
+	fd, err := c.OOMEventFD()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create oom event fd")
+	}
+	ec := make(chan Event)
+	go c.waitForOOMEvents(int(fd), ec)
+
+	return ec, nil
+}
+
+func (c *Manager) waitForOOMEvents(fd int, ec chan<- Event) {
+	for {
+		buffer := make([]byte, syscall.SizeofInotifyEvent*10)
+		bytesRead, err := syscall.Read(fd, buffer)
+		if err != nil {
+			log.Fatal(err)
+		}
+		var out map[string]interface{}
+		if bytesRead >= syscall.SizeofInotifyEvent {
+			if err := readStatsFile(c.path, "memory.events", out); err != nil {
+				val, ok := out["oom"]
+				if ok && val.(int) > oomEventCount {
+					oomEventCount = val.(int)
+					ec <- Event{}
+				}
+			}
+		}
+	}
 }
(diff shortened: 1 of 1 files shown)


v2/manager.go
```diff
@@ -411,3 +414,19 @@ func (c *Manager) freeze(path string, state State) error {
 		time.Sleep(1 * time.Millisecond)
 	}
 }
+
+func (c *Manager) OOMEventFD(rootPath string) (uintptr, error) {
```
AkihiroSuda: This should be called EventsFD, because the FD is not only for OOM

AkihiroSuda:
>  in case if we receive inotify event, create new separate file for oom's only, and return that file fd.

Does kernel have this feature? Seems not https://facebookmicrosites.github.io/cgroup2/docs/memory-controller.html

v2/manager.go
AkihiroSuda: We should also have function like
```go
const (
  EventLow
  EventHigh
  EventMax
  EventOOM
  EventOOMKill
)

func (*Manager) EventChan() (<-chan Event, error)
```

v2/manager.go
```diff
@@ -411,3 +414,19 @@ func (c *Manager) freeze(path string, state State) error {
...
+	fpath := filepath.Join(rootPath, "memory.events")
```
AkihiroSuda: you can use c.path

Zyqsempai:
@AkihiroSuda Do we actually care about other events here? Coz we can easily filter it only by `oom`

AkihiroSuda:
Yes, but I agree it is not now

Zyqsempai:
@AkihiroSuda You mean something like this? PTAL

v2/manager.go
```diff
@@ -36,10 +40,14 @@ const (
 	controllersFile = "cgroup.controllers"
 )
 
+var oomEventCount = 0
```
crosbymichael: lets remove the global

v2/manager.go
```diff
@@ -411,3 +419,50 @@ func (c *Manager) freeze(path string, state State) error {
 		time.Sleep(1 * time.Millisecond)
 	}
 }
+
+func (c *Manager) OOMEventFD() (uintptr, error) {
```
crosbymichael: I guess while we are working on this, can we go ahead and just make this generic and return all event types.  We can use the `Event` struct to pass the data back to users

v2/manager.go
```diff
@@ -411,3 +419,50 @@ func (c *Manager) freeze(path string, state State) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create oom event fd")
+	}
+	ec := make(chan Event)
+	go c.waitForOOMEvents(int(fd), ec)
+
+	return ec, nil
+}
+
+func (c *Manager) waitForOOMEvents(fd int, ec chan<- Event) {
```
AkihiroSuda: not specific to OOM

v2/manager.go
```diff
@@ -411,3 +419,50 @@ func (c *Manager) freeze(path string, state State) error {
...
+	for {
+		buffer := make([]byte, syscall.SizeofInotifyEvent*10)
+		bytesRead, err := syscall.Read(fd, buffer)
+		if err != nil {
+			log.Fatal(err)
```
AkihiroSuda: continue loop on error?

v2/manager.go
```diff
@@ -411,3 +419,50 @@ func (c *Manager) freeze(path string, state State) error {
...
+	}
+	defer syscall.InotifyRmWatch(fd, uint32(wd))
+
+	return uintptr(fd), nil
+}
+
+func (c *Manager) EventChan() (<-chan Event, error) {
+	fd, err := c.OOMEventFD()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create oom event fd")
```
AkihiroSuda: `errors.Errorf("failed to...")`

Zyqsempai:
@crosbymichael PTAL

crosbymichael:
rebase please :)

Zyqsempai:
@crosbymichael Done.

crosbymichael:
LGTM

PR #117, opened 2019-11-14
Added current and max swap and memory usage stats

Zyqsempai:
Related to #104 
Added `memory.current, memory.max, memory.swap.current, memory.swap.max` stats.


Zyqsempai:

estesp:
I think you will need to rebase on #118 when it is reviewed and merged to get CI passing for all your PRs.

v2/stats/metrics.proto
```diff
@@ -56,6 +56,15 @@ message MemoryStat {
...
 	uint64 thp_collapse_alloc = 31;
+	MemoryEntry usage = 32;
+	MemoryEntry swap_usage = 33;
+}
+
+message MemoryEntry {
+	uint64 limit = 1;
+	uint64 usage = 2;
+	uint64 max = 3;
+	uint64 failcnt = 4;
```
AkihiroSuda: we can omit unused fields

v2/stats/metrics.proto
```diff
@@ -56,6 +56,15 @@ message MemoryStat {
 	uint64 pglazyfreed = 29;
 	uint64 thp_fault_alloc = 30;
 	uint64 thp_collapse_alloc = 31;
+	MemoryEntry usage = 32;
```
AkihiroSuda: wondering this should be = 1

Zyqsempai: What do you mean?

crosbymichael: no

crosbymichael: instead of a struct, can we just add these as field? keep the stats flat and simple

v2/manager.go
```diff
@@ -326,6 +326,22 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 		ThpFaultAlloc:         out["thp_fault_alloc"].(uint64),
 		ThpCollapseAlloc:      out["thp_collapse_alloc"].(uint64),
 	}
+	cur_mem := getStatFileContent(filepath.Join(c.path, "memory.current"))
+	max_mem := getStatFileContent(filepath.Join(c.path, "memory.max"))
+	cur_swap := getStatFileContent(filepath.Join(c.path, "memory.swap.current"))
+	max_swap := getStatFileContent(filepath.Join(c.path, "memory.swap.max"))
+
+	usage := stats.MemoryEntry{
+		Usage: cur_mem,
```
crosbymichael: try not to use snake case `_` in the go code.

crosbymichael:
Can you rebase on master as well to fix the build?  Thanks!

v2/manager.go
Zyqsempai: oh, yeah, sorry, Python legacy;)

v2/stats/metrics.proto
Zyqsempai: Sure

Zyqsempai:
@crosbymichael Done.

v2/utils.go
```diff
@@ -216,3 +218,24 @@ func ToResources(spec *specs.LinuxResources) *Resources {
 	}
 	return &resources
 }
+
+// Gets uint64 parsed content of single value cgroup stat file
+func getStatFileContent(filePath string) uint64 {
```
fuweid: nit: function name is not specific, but it only can retrieve uint64 value.

AkihiroSuda: Also, this should take `io.Reader` rather than a filepath string and should have unit tests. Can be a follow-up PR.

Zyqsempai: Yep, i have in plans to cover everything with unit tests but in separate PR's

Zyqsempai: Also, I think that `io.Reader` parameter here will over complicate using of that function.

crosbymichael:
LGTM

PR #118, opened 2019-11-14
Set GO111MODULE=on and use `go mod vendor` to fix Go 1.11/1.12 builds

estesp:


estesp:
Seems like somewhere in the past 6 days using a mix of go.mod and `go get` started failing for non-Go 1.13 builds. Since we have a `go.mod` we might as well use it--switching to `go mod vendor` from `go get ..` makes the last 3 Golang releases work properly in Travis CI.

dqminh:
LGTM

@estesp its because we put the package inside GOPATH, and `go mod` doesnt work prior go 1.13 because the default `GO111MODULE=auto` will not active go mod inside a `GOPATH`.

Either one of these things would have make things work:
1. stop manually cloning some dependencies, put this outside of GOPATH, and do `go get`
2. forcing it to active module support with `GO111MODULE=on go get`. This is probably the same as `GO111MODULE=on go mod vendor` here.

I do prefer 1) for go modules project nowadays.

crosbymichael:
LGTM

PR #120, opened 2019-11-15
v2: fix issues

AkihiroSuda:
* pids stat was not functional due to the whitespace issue
* nil panic had happened when statting disabled controllers

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>


crosbymichael:
LGTM

PR #124, opened 2019-11-15
v2: toggle ancestors of c.path, not c.path itself

### LINKED ISSUE #121: v2: ToggleControllers() should toggle ancestors of c.path, not c.path itself
https://github.com/containerd/cgroups/blob/e56683d3d5a15abdea66d7511b5e16083da2f594/v2/manager.go#L192

`ToggleControllers()` should toggle ancestors of `c.path`, not `c.path` itself.

e.g. when `c.path` is like `/foo/bar/baz`, the following files need to be written:
* `/sys/fs/cgroup/foo/cgroup.subtree_control`
* `/sys/fs/cgroup/foo/bar/cgroup.subtree_control`

Note that `/sys/fs/cgroup/foo/bar/baz/cgroup.subtree_control` does not need to be written.

AkihiroSuda:
Fix #121

### DIFF
--- a/cmd/cgctl/main.go
+++ b/cmd/cgctl/main.go
@@ -46,6 +46,7 @@ func main() {
 		newCommand,
 		delCommand,
 		listCommand,
+		listControllersCommand,
 		statCommand,
 	}
 	app.Before = func(clix *cli.Context) error {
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

AkihiroSuda:
rebased

PR #127, opened 2019-11-20
Add Mode() func to check cgroup mount type

crosbymichael:
This function will return the cgroup mode that is mounted on the host system.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -25,13 +25,59 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	units "github.com/docker/go-units"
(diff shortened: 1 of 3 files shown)


utils.go
```diff
@@ -25,13 +25,59 @@ import (
...
+
+const (
+	// Legacy cgroups v1
+	Legacy CGMode = iota + 1
+	// Hybrid with cgroups v1 and v2 controllers mounted
+	Hybrid
+	// Unified with only cgroups v2 mounted
+	Unified
+	// Unavailable cgroup mountpoint
+	Unavailable
```
AkihiroSuda: =iota?

PR #129, opened 2019-11-22
Added v2 memory test

Zyqsempai:
@AkihiroSuda @crosbymichael PTAL, what do you think, can we move in that direction with unit tests? or you saw it in a different way?

### DIFF
--- a/v2/memoryv2_test.go
+++ b/v2/memoryv2_test.go
@@ -35,7 +35,7 @@ func TestCgroupv2MemoryStats(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	defer clear(groupPath)
+	defer os.RemoveAll(groupPath)
 	res := Resources{
 		CPU:  &CPU{},
 		Pids: &Pids{},
@@ -65,7 +65,7 @@ func TestCgroupv2MemoryStats(t *testing.T) {
 	assert.Equal(t, uint64(629145600), stats.Memory.UsageLimit)
 	swapMax, err := ioutil.ReadFile(filepath.Join(c.path, "memory.swap.max"))
 	if err != nil {
-		t.Fatal("failed to read memory.max file: ", err)
+		t.Fatal("failed to read memory.swap.max file: ", err)
 	}
 	assert.Equal(t, "314572800", strings.TrimSpace(string(swapMax)))
 
@@ -74,8 +74,4 @@ func TestCgroupv2MemoryStats(t *testing.T) {
 		t.Fatal("failed to read memory.max file: ", err)
 	}
 	assert.Equal(t, "629145600", strings.TrimSpace(string(memMax)))
-}
-
-func clear(path string) {
-	os.RemoveAll(path)
 }
--- a/v2/memoryv2_test.go
+++ b/v2/memoryv2_test.go
@@ -17,6 +17,7 @@
 package v2
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
I think we should use real cgroup for testing

Zyqsempai:
@AkihiroSuda You mean something like this? Check it out please, I updated code a little bit.

Zyqsempai:
@AkihiroSuda I tried few approaches with `real cgroup` and every time struggling with permissions. I didn't find in any realization of cgroups real cgroup usage for tests.
How do you plan to do so? Coz I have no more ideas(I tried to mount it, tried to use existing one in /sys/fs/cgroup/unified)

v2/testutils_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+const defaultCgroup2Path = "/sys/fs/cgroup/unified"
```
AkihiroSuda: Use `/sys/fs/cgroup`. You can just skip  the test when running in hybrid mode.

v2/memoryv2_test.go
```diff
@@ -0,0 +1,59 @@
...
+			Max:  pointerInt64(629145600),
+			Swap: pointerInt64(314572800),
+			High: pointerInt64(524288000),
+		},
+	}
+	c, err := NewManager(cg.cgv2Root, "/memory-test-cg", &res)
+	if err != nil {
+		t.Fatal("failed to init new cgroup manager: ", err)
+	}
+	cg.writeFileContents(map[string]string{
```
AkihiroSuda: toggle the controller and write the limit files via the manager, and then verify the files via `ioutil.ReadFile`, and then also verify the stat struct from the manager.

v2/memoryv2_test.go
```diff
@@ -0,0 +1,73 @@
...
+		Pids: &Pids{},
+		IO:   &IO{},
+		RDMA: &RDMA{},
+		Memory: &Memory{
+			Max:  pointerInt64(629145600),
+			Swap: pointerInt64(314572800),
+			High: pointerInt64(524288000),
+		},
+	}
+	c, err := NewManager(cg.cgv2Root, cg.groupPath, &res)
```
AkihiroSuda: `NewManager` should take `/sys/fs/cgroup` as the root

Zyqsempai:

v2/memoryv2_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCgroupv2MemoryStats(t *testing.T) {
+	cg := NewCgroupDir("/memory-test-cg", t)
```
AkihiroSuda: I don't think you need this helper function

AkihiroSuda: Other parts looks good

v2/memoryv2_test.go
```diff
@@ -0,0 +1,81 @@
...
+	assert.Equal(t, "314572800", strings.TrimSpace(string(swapMax)))
+
+	memMax, err := ioutil.ReadFile(filepath.Join(c.path, "memory.max"))
+	if err != nil {
+		t.Fatal("failed to read memory.max file: ", err)
+	}
+	assert.Equal(t, "629145600", strings.TrimSpace(string(memMax)))
+}
+
+func clear(path string) {
```
fuweid: nit: it looks like we can call `os.RemoveAll(path` directly, instead of using helper function.

Zyqsempai:
@fuweid Fixed.

crosbymichael:
LGTM

Issue #128, opened 2019-11-20
v2: mem.Reservation conversion is wrong


AkihiroSuda:
https://github.com/containerd/cgroups/blob/9e089962ffffaeb9281c22c14bdb77b59852f7a1/v2/utils.go#L191

This should be converted into `memory.low`, not `memory.high`
ref: https://github.com/containers/crun/pull/188/files

Zyqsempai:
I will fix it!

PR #130, opened 2019-11-25
Fixed memory conversion for `reservation` from high to low.

Zyqsempai:
Fix: #128 
Changed memory conversion for `reservation` from high to low. 

### DIFF
--- a/v2/memory.go
+++ b/v2/memory.go
@@ -20,6 +20,7 @@ type Memory struct {
 	Swap *int64
 	Max  *int64
 	Low  *int64
+	High *int64
 }
 
 func (r *Memory) Values() (o []Value) {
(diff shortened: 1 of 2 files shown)


Zyqsempai:
@AkihiroSuda PTAL

v2/memory.go
```diff
@@ -19,7 +19,7 @@ package v2
 type Memory struct {
 	Swap *int64
 	Max  *int64
-	High *int64
```
AkihiroSuda: no need to remove High?

Zyqsempai: Is this a question? coz I removed it;)

Zyqsempai: Or you wanted to say that we still need `High`? If so, then what for? what we will store there?

AkihiroSuda: Probably we don't have usecase atm, but the package should be able to handle all features supported by the kernel.

Zyqsempai: Ok, I will put it back then.

Zyqsempai: Done

PR #134, opened 2019-12-17
Added period configuration for cpu.max

### LINKED ISSUE #133: Set $PERIOD in cpu.max
In documentation we have next description for `cpu.max`
```
cpu.max
	A read-write two value file which exists on non-root cgroups.
	The default is "max 100000".

	The maximum bandwidth limit.  It's in the following format::

	  $MAX $PERIOD

	which indicates that the group may consume upto $MAX in each
	$PERIOD duration.  "max" for $MAX indicates no limit.  If only
	one number is written, $MAX is updated.
```
Currently we updating only $MAX, do we care about $PERIOD at all? If so I will open corresponding PR.

Zyqsempai:
Fixes #133 

Added quota parameter to `cpu.max` resource control.
Added test for `ToResources` function.

### DIFF
--- a/v2/cpu.go
+++ b/v2/cpu.go
@@ -18,7 +18,7 @@ package v2
 
 type CPU struct {
 	Weight *uint64
-	Max    *uint64
+	Max    string
 	Cpus   string
 	Mems   string
 }
@@ -30,10 +30,10 @@ func (r *CPU) Values() (o []Value) {
 			value:    *r.Weight,
 		})
 	}
-	if r.Max != nil {
+	if r.Max != "" {
 		o = append(o, Value{
 			filename: "cpu.max",
-			value:    *r.Max,
+			value:    r.Max,
 		})
 	}
 	if r.Cpus != "" {
--- a/v2/cpu.go
+++ b/v2/cpu.go
@@ -16,9 +16,24 @@
 
 package v2
 
+import (
+	"strconv"
+	"strings"
+)
+
+type CPUMax string
+
+func NewCPUMax(quota *int64, period *uint64) CPUMax {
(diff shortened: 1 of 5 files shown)


v2/cpu.go
```diff
@@ -18,7 +18,7 @@ package v2
 
 type CPU struct {
 	Weight *uint64
-	Max    *uint64
+	Max    string
```
crosbymichael: @AkihiroSuda do we want to handle "max" this way or something else?  a bool value or two fields from max and period?

AkihiroSuda: I prefer having two fields

crosbymichael: same, these should be two separate fields

Zyqsempai: and how do you plan to handle it in next steps? I mean writing to the file, this will be the unique case and wanted to make it in standard way.

Zyqsempai: Guys, question I still actual, we have this func https://github.com/containerd/cgroups/blob/master/v2/manager.go#L116
I don't prefer to add exceptional case here. And since anyway we will join those two fields values in the end, I don't see a point to add two separate fields.

AkihiroSuda: Maybe like this? 
```go
type CPUMax string
func NewCPUMax(max uint64, perod *uint64) CPUMax { }
type CPU struct {
  Weight *uint64
  Max    CPUMax
  ...
}
```

Zyqsempai: Did you me something like this? PTAL on last commit

Zyqsempai:
@crosbymichael Can we merge it? I need this thingy to finish `systemd` manager

PR #145, opened 2020-02-05
memory.events handling rework

Zyqsempai:
Small rework of the memory.events handler in order to implement OOM handler in main containerD repo.

During the memory.events handler implementations, were made few mistakes, this PR fixes them.

### DIFF
--- a/v2/manager.go
+++ b/v2/manager.go
@@ -526,6 +526,7 @@ func (c *Manager) freeze(path string, state State) error {
 	}
 }
 
+// MemoryEventFD returns inotify file descriptor and 'memory.events' inotify watch descriptor
 func (c *Manager) MemoryEventFD() (int, uint32, error) {
 	fpath := filepath.Join(c.path, "memory.events")
 	fd, err := syscall.InotifyInit()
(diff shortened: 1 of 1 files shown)


Zyqsempai:

v2/manager.go
```diff
@@ -526,20 +526,19 @@ func (c *Manager) freeze(path string, state State) error {
 	}
 }
 
-func (c *Manager) MemoryEventFD() (uintptr, error) {
+func (c *Manager) MemoryEventFD() (int, uint32, error) {
```
AkihiroSuda: needs godoc

Zyqsempai:
@AkihiroSuda Added goDoc and updated PR description, related to "how to test it" i tried to add test but I wasn't able to properly simulate OOM behavior for process in the test.
Maybe you have some suggestions, how to test it properly, coz I am more than agree that we need few tests here.

Zyqsempai:
@AkihiroSuda Can we move forward this one, coz implementation of the OOM handler in containerD main repo depends on this one.

AkihiroSuda:
@fuweid PTAL?

v2/manager.go
AkihiroSuda: > (int, uint32

why different types?

Zyqsempai: Coz this is the base types for file descriptor and watch descriptor, at least `syscall.InotifyRmWatch()` expects it in that way.

PR #147, opened 2020-03-05
throttle.* metrics must be kept for non-CFQ schedulers

kzys:
Since c4b9ac5, Stat() ignores
/sys/fs/cgroup/blkio/$cgroup/blkio.throttle.io_serviced and
/sys/fs/cgroup/blkio/$cgroup/blkio.throttle.io_service_bytes when
/sys/fs/cgroup/blkio/$cgroup/blkio.io_serviced_recursive is there.

However the existance of this file only indicates that the kernel is
compiled with the CFQ scheduler. There is a chance that a cgroup uses
devices with a different scheduler such as "none".

### DIFF
--- a/blkio.go
+++ b/blkio.go
@@ -90,54 +90,47 @@ func (b *blkioController) Update(path string, resources *specs.LinuxResources) e
 
 func (b *blkioController) Stat(path string, stats *v1.Metrics) error {
 	stats.Blkio = &v1.BlkIOStat{}
-	settings := []blkioStatSettings{
-		{
-			name:  "throttle.io_serviced",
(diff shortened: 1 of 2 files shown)


kzys:
Looks like the test failure is not related to my change.

```
Timed out while waiting for the machine to boot. This means that

Vagrant was unable to communicate with the guest machine within

the configured ("config.vm.boot_timeout" value) time period.
```

kzys:
@Zyqsempai Thanks! The test is passing, while I have to skip the test in Travis though.

AkihiroSuda:
@fuweid PTAL

crosbymichael:
LGTM

PR #151, opened 2020-03-12
bufio.Scanner.Err nits; mountinfo parser fixes

kolyshkin:
### bufio.Scanner.Err nits

1. The call to `Err()` should be done after the `Scan()` loop,
       not inside it.
    
2. Add the call to `Err()` to after the `Scan()` loop, otherwise
       we might miss the "read only half of it" kind of error.
####    getCgroupDestination: speedup and improve
    
1. Use `strings.Split` instead of `strings.Fields`. Split is 2x faster!
    
2. Check the number of fields and fs type before parsing options,
       this should speed things up a bit, and prevent a potential
       panic caused by invalid slice indexes.

####     v1MountPoint: simplify/speedup a bit

* Instead of relying on one split + a linear search for 
  separator + another split, let's assume the last
  three fields are the ones that are post-sepatator.

### DIFF
--- a/blkio.go
+++ b/blkio.go
@@ -181,9 +181,6 @@ func (b *blkioController) readEntry(devices map[deviceKey]string, path, name str
 	defer f.Close()
 	sc := bufio.NewScanner(f)
 	for sc.Scan() {
-		if err := sc.Err(); err != nil {
-			return err
-		}
 		// format: dev type amount
 		fields := strings.FieldsFunc(sc.Text(), splitBlkIOStatLine)
 		if len(fields) < 3 {
@@ -220,7 +217,7 @@ func (b *blkioController) readEntry(devices map[deviceKey]string, path, name str
 			Value:  v,
 		})
 	}
-	return nil
+	return sc.Err()
 }
 
 func createBlkioSettings(blkio *specs.LinuxBlockIO) []blkioSettings {
--- a/utils.go
+++ b/utils.go
@@ -320,7 +320,14 @@ func getCgroupDestination(subsystem string) (string, error) {
 	defer f.Close()
 	s := bufio.NewScanner(f)
 	for s.Scan() {
-		fields := strings.Fields(s.Text())
+		fields := strings.Split(s.Text(), " ")
+		if len(fields) < 10 {
+			// broken mountinfo?
+			continue
+		}
+		if fields[len(fields)-3] != "cgroup" {
+			continue
+		}
 		for _, opt := range strings.Split(fields[len(fields)-1], ",") {
 			if opt == subsystem {
 				return fields[3], nil
--- a/utils.go
+++ b/utils.go
@@ -181,6 +181,10 @@ func readPids(path string, subsystem Name) ([]Process, error) {
 			})
 		}
 	}
+	if err := s.Err(); err != nil {
+		// failed to read all pids?
+		return nil, err
+	}
 	return out, nil
 }
 
@@ -207,6 +211,9 @@ func readTasksPids(path string, subsystem Name) ([]Task, error) {
 				Path:      path,
 			})
 		}
+	}
+	if err := s.Err(); err != nil {
+		return nil, err
 	}
 	return out, nil
 }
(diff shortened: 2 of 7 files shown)


blkio.go
```diff
@@ -181,9 +181,6 @@ func (b *blkioController) readEntry(devices map[deviceKey]string, path, name str
 	defer f.Close()
 	sc := bufio.NewScanner(f)
 	for sc.Scan() {
-		if err := sc.Err(); err != nil {
-			return err
-		}
```
Zyqsempai: I didn't get, so you suggest to ignore scan errors and continue to try to collect data from broken scan? Ho should it work?

kolyshkin: Please see https://golang.org/pkg/bufio/#Scanner.Scan documentation. In case of an error, Scan() returns false, and the Err() should be checked right after the  `for s.Scan()` loop.

kolyshkin:
> Please also, squash your commits

Are you suggesting to put unrelated changes into a single commit? I don't think it's a good idea.

> and sign the final one.

All my commits are signed.

utils.go
```diff
@@ -320,7 +320,14 @@ func getCgroupDestination(subsystem string) (string, error) {
 	defer f.Close()
 	s := bufio.NewScanner(f)
 	for s.Scan() {
-		fields := strings.Fields(s.Text())
+		fields := strings.Split(s.Text(), " ")
+		if len(fields) < 10 {
+			// broken mountinfo?
+			continue
+		}
+		if fields[len(fields)-3] != "cgroup" {
```
kolyshkin: This assumes we do not use this function for cgroupv2 mount (as far as I can see we're not, since the v2 unified mount point does not list individual controllers in mount options).

v1.go
```diff
@@ -54,28 +54,20 @@ func v1MountPoint() (string, error) {
...
-			postSeparatorFields = strings.Fields(text[index+3:])
-			numPostFields       = len(postSeparatorFields)
+			text      = scanner.Text()
+			fields    = strings.Split(text, " ")
+			numFields = len(fields)
 		)
-		// this is an error as we can't detect if the mount is for "cgroup"
-		if numPostFields == 0 {
-			return "", fmt.Errorf("Found no fields post '-' in %q", text)
+		if numFields < 10 {
```
crosbymichael: i wonder if this is going to break anything, i remember some kernels have in the best, returned various numbers in the mount outputs

AkihiroSuda:
@kolyshkin Do you plan to use `moby/sys/mount` in this PR? Or will it be a separate PR?

crosbymichael:
We don't want moby deps

v1.go
kolyshkin: I am aware of two kernel bugs:

1. empty "source" field. this is not affecting this parser because it uses strings.Split by a space, and the extra space is there (for details, see https://github.com/moby/sys/pull/5).

2. extra fields added by cifs mounts with `unc=` option with value containing spaces. This makes numFileds greater, not less.

IOW I am not aware of anything that would break this condition.

crosbymichael: ok, thanks.  just wanted to double check

kolyshkin:
> We don't want moby deps

In fact I hope you would consider it. Not moby/moby, but https://github.com/moby/sys (which only has mount and mountinfo, but I plan to add xattr as well).

I spent most of last week looking into the unfortunate results of copy-pasting the mountinfo parsing code between the projects. As a result of that copy-pasting, we have 5+ implementations of mountinfo parser (in runc, containerd, kubernetes, cilium, etc). Most of those implementations are derived from moby (github.com/docker/docker/pkg/mount) and they all carry the same issues (mostly performance).

Some of those issues were fixed in moby/moby, but those changes did not made its way to all the projects that borrowed the docker implementation. As a result, we need to fix this in many places instead of one. This is the technical debt I want to address, by creating https://github.com/moby/sys  and having a good implementation of code everyone borrows. The plan is also to have minimal dependencies (for now this is standard go library and `golang.org/x/sys`, and I hope it will stay that way).

Now, if you don't like anything with the moby word in it, I am very open to moving the above repo to say under github.com/opencontainers/, just let me know.

kolyshkin:
> @kolyshkin Do you plan to use moby/sys/mount in this PR? Or will it be a separate PR?

I'd love to, yes, but we can address it in a separate PR, this one looks good per se.

crosbymichael:
@kolyshkin ok, i'll look at that sys repo.  We just didn't want things from the main `moby/moby` repo as it's a large dep and can change often

crosbymichael:
LGTM

PR #154, opened 2020-03-30
Implement registering for other memory notifications besides oom

dcantah:
* Implement functionality to register for events on the other memory cgroup files
that support the notification API (memory.usage_in_bytes, memory.memsw.usage_in_bytes,
memory.pressure_limit).

* Small typo fix in hierarchy.go.  enableds -> enables

Signed-off-by: Daniel Canter <dcanter@microsoft.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,27 @@ err := control.MoveTo(destination)
 subCgroup, err := control.New("child", resources)
 ```
 
+### Registering for memory events
+
+This allows you to get notified by an eventfd for v1 memory cgroups events.
+
+```go
+event := cgroups.MemoryThresholdEvent(50 * 1024 * 1024, false)
+efd, err := control.RegisterMemoryEvent(event)
+```
+
+```go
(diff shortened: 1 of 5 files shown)


crosbymichael:
What do you think about a single method that can return `uintptr, error` but handle args and the type of event that you would like to receive?  Just thinking so we could keep the interface small and support these and any other events from the subsystems in the future.

dcantah:
@crosbymichael I agree this would be better but I didn't know if you guys wanted to keep OOMEventFD as part of the interface so I simply added the new events to it as well. In your proposal we would be getting rid of OOMEventFD and the two new ones in favor of just one method (let's call it MemoryEvent or similar) if I'm understanding correctly?

crosbymichael:
I think we would need to keep `OOMEventFD` for a little bit to give ppl time to move over to the new method.

`MemoryEvent` sounds good to me.

dcantah:
@crosbymichael Cool I'll get to work

crosbymichael:
thanks!

dcantah:
@crosbymichael PTAL

crosbymichael:
LGTM

PR #158, opened 2020-05-20
v2: fix EventChan

AkihiroSuda:
EventChan() was completely broken:
* [critical] `err == nil` comparison was flipped in the opposite way
* [critical] `var out map[string]interface{}` was not initialized with `make()`
* [non-critical] `.(uint64)` conversin errors were not caught


AkihiroSuda:
@Zyqsempai @fuweid PTAL

AkihiroSuda:
rebased

PR #150, opened 2020-03-11
Added support for memory.events stats

Zyqsempai:
Fixes #149 
Added support for memory.events stats.

Signed-off-by: Boris Popovschi <zyqsempai@mail.ru>

PR #160, opened 2020-05-20
[Carry #150] Added support for memory.events stats

### LINKED ISSUE #149: cgroup2: support statting memory.events*
The PR for `docker stats` (https://github.com/moby/moby/pull/40657) require `containerd/cgroups` support for `memory.events*`.

**EDIT**: What we can add now seems only `FailCnt`, and this is not urgent (not used by `docker` CLI)

AkihiroSuda:
Carry #150 
Fix #149


v2/stats/metrics.proto
```diff
@@ -63,6 +64,12 @@ message MemoryStat {
 	uint64 usage_limit = 33;
 	uint64 swap_usage = 34;
 	uint64 swap_limit = 35;
+    // memory.events (101-200)
```
AkihiroSuda, line 67: RFC: Is this numbering scheme right way?
cc @stevvooe

crosbymichael: Why wouldn't you just add them in order....???

AkihiroSuda: Because `memory.stat` will eventually have more fields.

v2/manager.go
```diff
@@ -356,6 +356,7 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 		return nil, err
 	}
 	out := make(map[string]interface{})
+	memoryEvents := make(map[string]interface{})
```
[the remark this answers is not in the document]
AkihiroSuda: done

v2/stats/metrics.proto
crosbymichael: but does the proto numbers really matter that much to us that we have to put an upper limit on the memory fields?  Why not just add them after the events?

AkihiroSuda: added new message struct `MemoryEvents`

AkihiroSuda:
updated proto

v2/manager.go
```diff
@@ -427,7 +433,13 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 		SwapUsage:             getStatFileContentUint64(filepath.Join(c.path, "memory.swap.current")),
 		SwapLimit:             getStatFileContentUint64(filepath.Join(c.path, "memory.swap.max")),
 	}
-
+	metrics.MemoryEvents = &stats.MemoryEvents{
```
fuweid, line 436: if map `memoryEvents` is empty, do we just let `metrics.MemoryEvents` nil here?

AkihiroSuda: updated

AkihiroSuda:
@mikebrow Still LGTY?

Issue #164, opened 2020-06-08
DCO check isn't working


AkihiroSuda:
We should have DCO checking like this: 
https://github.com/containerd/go-cni/blob/0553354f0046ccd41a02e724826040491a3d8998/.travis.yml#L18

estesp:
Looks like adding the matrix entry that does the v2 test via travis+vagrant is somehow skipping the existing "`script:`" section which already had a working DCO check.

I might take a stab at switching this repo over to GH actions since you've done the work to get Vagrant up under GH actions+MacOS instance.

PR #165, opened 2020-06-08
Migrate CI to GitHub Actions

estesp:
Since the `Vagrantfile` was also being used by Travis this PR removes Travis at the same time as adding Actions, given Travis would simply fail given the changes to the Vagrantfile which are aligned with using the MacOS instance in GH Actions.

You can see a successful run of these changes in my fork [here](https://github.com/estesp/cgroups/runs/751664439?check_suite_focus=true)

Fixes: #164 (which was a result of the matrix setup in Travis overriding the standard script section--GH Actions allows us to have a common project check area for DCO and headers, and then run 2 dependent jobs after that for v1 and v2 respectively)


estesp:
Note that no CI will run on this PR because Travis's config file is removed, and actions only run in the fork, so until this merged, PRs won't run the actions defined here within the PR as a security measure. See the link above to see the output from these exact changes in my fork.

Vagrantfile
```diff
@@ -13,13 +13,34 @@ Vagrant.configure("2") do |config|
...
-config exclude kernel,kernel-core
-config install_weak_deps false
-update
-install git golang-go
-ts run
+    set -eux -o pipefail
+    # configuration
+    GO_VERSION="1.13.11"
+
+	# install gcc and Golang
```
AkihiroSuda, line 20: nit: indentation

estesp: thanks! fixed

estesp:
2 LGTMs so merging to enable full CI again in this repo. Thanks!

estesp:
Merge to master run shows actions CI now active: https://github.com/containerd/cgroups/runs/754016151

PR #175, opened 2020-07-22
memory: remove wrong memory.kmem.limit_in_bytes check

Snorch:
This check is useless as:

"We have to limit the kernel memory here as it won't be accounted at all
until a limit is set on the cgroup" - if kmem accounting is enabled on
boot (no nokmem) kernel memory would be accounted without setting the
limit.

"limit cannot be set once the cgroup has children, or if there are
already tasks in the cgroup." - It can, the only restriction it should
be > usage, but setting this limit < future usage is a bad thing itself.
So there is no difference when to set this limit.

More over memory.kmem.limit_in_bytes is deprecated in mainstream linux
kernel as it was considered unreliable, see more information in:
    
commit 0158115f702b ("memcg, kmem: deprecate kmem.limit_in_bytes")
Link: https://github.com/torvalds/linux/commit/0158115f702b

Signed-off-by: Pavel Tikhomirov <ptikhomirov@virtuozzo.com>

### DIFF
--- a/memory.go
+++ b/memory.go
@@ -24,7 +24,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"syscall"
 
 	v1 "github.com/containerd/cgroups/stats/v1"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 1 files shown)


Snorch:
Are there any news? Can it be merged?

AkihiroSuda:
Sorry needs one more LGTM. Probably after v1.4 (soon)

AkihiroSuda:
@estesp @fuweid PTAL

Snorch:
Thanks!

PR #183, opened 2020-11-09
utils: export RunningInUserNS()

thaJeztah:
Exporting this utility, so that consumers of the containerd/cgroups package can use this, without having to duplicate this functionality, for example; https://github.com/containerd/containerd/blob/0088c2de8019dfd38040634931ce05e04b1aa74f/sys/userns_linux.go#L31-L62

The only difference (currently) is that containerd/cgroups does not have a "stub" for this function for non-Linux (whereas containerd/sys does); not sure if it would make sense to add a stub here (given that this repository is assuming Linux)

### DIFF
--- a/subsystem.go
+++ b/subsystem.go
@@ -59,7 +59,7 @@ func Subsystems() []Name {
 		Blkio,
 		Rdma,
 	}
-	if !isUserNS {
+	if !RunningInUserNS() {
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

estesp (approved):
LGTM

PR #190, opened 2021-02-03
Added comments explaining why all Memory___Event functions return a MemoryEvent interface

AlvinKuruvilla:


dcantah:
Sorry for the delay here, I believe the CI is failing due to the commit not being signed off on. Can do `git commit --amend --sign-off`

memory.go
```diff
@@ -45,6 +45,7 @@ type memoryThresholdEvent struct {
 
 // MemoryThresholdEvent returns a new memory threshold event to be used with RegisterMemoryEvent.
 // If swap is true, the event will be registered using memory.memsw.usage_in_bytes
+//MemoryThresholdEvent returns a MemoryEvent interface so they can all be registered via the memory controllers memoryEvent method
```
dcantah, line 48: Reading this more, I think the first line of the comment (for both) should be changed to say it returns a `MemoryEvent` interface and then a subsequent line can mention what the event actually does/is registered for ("This event will trigger on the threshold being reached etc etc")

kzys:
@AlvinKuruvilla Let me know if you are still working on this PR.

AlvinKuruvilla:
Sorry for the late reply. I have a lot on my plate at the moment so I don't think I will have a chance to get around to this PR. Sorry about that

PR #195, opened 2021-05-05
Use /proc/partitions to get device names

SaveTheRbtz:
On systems with hundreds of disks/CPUs [/proc/partitions](https://elixir.bootlin.com/linux/v5.12.1/source/block/genhd.c#L935) is way less expensive than [/proc/diskstats](https://elixir.bootlin.com/linux/v5.12.1/source/block/genhd.c#L1265).

Here is a `perf top` example from a system that has hundreds of `loop` mounts:
Screen Shot 2021-05-04 at 8 36 25 PM

Since the code is only using major/minor numbers and drive name we should probably use a way more light-weight `/proc/partitions` endpoint

Signed-off-by: Alexey Ivanov <rbtz@dropbox.com>

### DIFF
--- a/blkio_test.go
+++ b/blkio_test.go
@@ -24,18 +24,19 @@ import (
 	v1 "github.com/containerd/cgroups/stats/v1"
 )
 
-const data = `   7       0 loop0 0 0 0 0 0 0 0 0 0 0 0
-   7       1 loop1 0 0 0 0 0 0 0 0 0 0 0
-   7       2 loop2 0 0 0 0 0 0 0 0 0 0 0
-   7       3 loop3 0 0 0 0 0 0 0 0 0 0 0
-   7       4 loop4 0 0 0 0 0 0 0 0 0 0 0
-   7       5 loop5 0 0 0 0 0 0 0 0 0 0 0
-   7       6 loop6 0 0 0 0 0 0 0 0 0 0 0
-   7       7 loop7 0 0 0 0 0 0 0 0 0 0 0
-   8       0 sda 1892042 187697 63489222 1246284 1389086 2887005 134903104 11390608 1 1068060 12692228
-   8       1 sda1 1762875 37086 61241570 1200512 1270037 2444415 131214808 11152764 1 882624 12409308
-   8       2 sda2 2 0 4 0 0 0 0 0 0 0 0
-   8       5 sda5 129102 150611 2244440 45716 18447 442590 3688296 67268 0 62584 112984`
+const data = `major minor  #blocks  name
+
+   7        0          4 loop0
+   7        1     163456 loop1
+   7        2     149616 loop2
+   7        3     147684 loop3
+   7        4     122572 loop4
+   7        5       8936 loop5
+   7        6      31464 loop6
+   7        7     182432 loop7
+ 259        0  937692504 nvme0n1
+ 259        1      31744 nvme0n1p1
+`
 
 func TestGetDevices(t *testing.T) {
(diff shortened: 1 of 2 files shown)


Zyqsempai (approved):
I like that, @AkihiroSuda what about you?

blkio_test.go
```diff
@@ -24,32 +24,38 @@ import (
...
-   7       4 loop4 0 0 0 0 0 0 0 0 0 0 0
-   7       5 loop5 0 0 0 0 0 0 0 0 0 0 0
-   7       6 loop6 0 0 0 0 0 0 0 0 0 0 0
-   7       7 loop7 0 0 0 0 0 0 0 0 0 0 0
-   8       0 sda 1892042 187697 63489222 1246284 1389086 2887005 134903104 11390608 1 1068060 12692228
-   8       1 sda1 1762875 37086 61241570 1200512 1270037 2444415 131214808 11152764 1 882624 12409308
-   8       2 sda2 2 0 4 0 0 0 0 0 0 0 0
-   8       5 sda5 129102 150611 2244440 45716 18447 442590 3688296 67268 0 62584 112984`
+const data = `major minor  #blocks  name
+
```
AkihiroSuda, line 28: Is this line guaranteed to be an empty? Is there a kernel documentation about this?

SaveTheRbtz: So, the kernel documentation on the subject [is rather scarce](https://man7.org/linux/man-pages/man5/proc.5.html).  In the code 2 lines are guaranteed and were there for quite a while (at least from [2.6.32](https://elixir.bootlin.com/linux/v2.6.32/source/block/genhd.c#L733)).  

PS. That said, [/proc/diskstats](https://elixir.bootlin.com/linux/v5.12.1/source/block/genhd.c#L1256) has eventually disabled this header.  I'm not sure we want to complicate the code just in case they ever do that for /proc/partitions.

AkihiroSuda (approved):

estesp:
Can you squash the commits? Only the first one has a DCO sign-off.

SaveTheRbtz:
@Zyqsempai @estesp @AkihiroSuda Friendly ping x2 =))

estesp (approved):
LGTM

SaveTheRbtz:
Seems like the test fails due to the timeout.  I do not see anything in the logs:
```
PASS
ok  	github.com/containerd/cgroups/v2	0.084s
?   	github.com/containerd/cgroups/v2/stats	[no test files]
Error: The operation was canceled.
```

Is that a real problem or a CI configuration issue?

estesp:
Looks like vagrant-based cgroup v2 tests are now passing the 20m limit and are getting canceled. I just opened #196 to see if 30 minutes is sufficient.

estesp:
Is it possible your change is causing a hang in some test under vagrant or a cgroup "v2 mount" booted system? My tests ran in a few minutes in #196 without your change.

SaveTheRbtz:
Interesting, i would say that this should be quite unlikely, given that:
* vagrant tests stall after they report success (`PASS`).
* they pass on vagrant that I run locally.

Can we try a double experiment, just in case:
* can you rebase your change in #196 on top of this diff to check that it will start failing.
* remove the rebase and verify that it passes again.
* (optional) kick off tests here once more to check for any transient CI problems.

estesp:
Looks like it was some form of transient failure in CI. Thanks!

SaveTheRbtz:
@estesp is it possible to cut a release so other projects can start picking this change?

PR #198, opened 2021-06-09
allow build time opt-out of systemd dependencies

metux:
Allow building without systemd specific code and its dependencies by
setting the 'no_systemd' flag.

Signed-off-by: Enrico Weigelt, metux IT consult <info@metux.net>

### DIFF
--- a/cmd/cgctl/main_nosystemd.go
+++ b/cmd/cgctl/main_nosystemd.go
@@ -1,0 +1,39 @@
+// +build no_systemd
+
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/urfave/cli"
+)
+
+var newSystemdCommand = cli.Command{
+	Name:  "systemd",
+	Usage: "(disabled)",
+	Action: func(clix *cli.Context) error {
+		return nil
+	},
+}
+
+var deleteSystemdCommand = cli.Command{
+	Name:  "del-systemd",
+	Usage: "(disabled)",
+	Action: func(clix *cli.Context) error {
+		return nil
+	},
+}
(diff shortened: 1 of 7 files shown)


cmd/cgctl/main_nosystemd.go
```diff
@@ -0,0 +1,39 @@
...
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"github.com/urfave/cli"
+)
+
+var newSystemdCommand = cli.Command{
```
kzys, line 25: Do we want to keep these subcommands with "(disabled)" or just remove subcommands from the build?

kzys: For example, curl's usage message is pretty much unaware about build-time options.

I built this curl with `--disable-ftp`. The help message is there, but it is not supported.

```
[ec2-user@ip-172-31-25-68 curl]$ ./src/curl --help ftp
Usage: curl [options...] <url>
ftp: FTP protocol options
 -a, --append        Append to target file when uploading
     --crlf          Convert LF to CRLF in upload
     --disable-eprt  Inhibit using EPRT or LPRT
     --disable-epsv  Inhibit using EPSV
 -D, --dump-header <filename>  Write the received headers to <filename>
     --ftp-account <data>  Account data string
     --ftp-alternative-to-user <command>  String to replace USER [name]
     --ftp-create-dirs  Create the remote dirs if not present
     --ftp-method <method>  Control CWD usage
     --ftp-pasv      Use PASV/EPSV instead of PORT
 -P, --ftp-port <address>  Use PORT instead of PASV
     --ftp-pret      Send PRET before PASV
     --ftp-skip-pasv-ip  Skip the IP address for PASV
     --ftp-ssl-ccc   Send CCC after authenticating
     --ftp-ssl-ccc-mode <active/passive>  Set CCC mode
     --ftp-ssl-control  Require SSL/TLS for FTP login, clear for transfer
 -I, --head          Show document info only
     --ignore-content-length  Ignore the size of the remote resource
     --krb <level>   Enable Kerberos with security <level>
 -l, --list-only     List only mode
 -Q, --quote         Send command(s) to server before transfer
 -r, --range <range>  Retrieve only the bytes within RANGE
 -z, --time-cond <time>  Transfer based on a time condition
[ec2-user@ip-172-31-25-68 curl]$ ./src/curl ftp://localhost/foobar
curl: (1) Protocol "ftp" not supported or disabled in libcurl
[ec2-user@ip-172-31-25-68 curl]$ 
```

samuelkarp: I'd keep them and make the action return an error if they're used, rather than doing nothing.

kzys: Returning non-zero exit code makes sense for me and more Unix-like. @metux What do you think?

kzys:
Closing due to inactivity.

Compared to https://github.com/containerd/containerd/pull/5591, I'm on the fence regarding making systemd optional in containerd/cgroups since it is a low-level package.

PR #201, opened 2021-07-15
v2: remove unimplemented errors and ErrorHandler, IgnoreNotExist

thaJeztah:
relates to https://github.com/containerd/containerd/pull/5739

### v2: remove errors that are never returned

Unlike v1, the v2 package never returns these errors. This patch removes
them to prevent consumers of this package from assuming they may be returned,
and writing special handling around them.

> **NOTE: this is a breaking change, so (following SemVer) would require a major
> version update**. However, consumers should not use these errors. Alternatively,
> we could "deprecate" them first. Making it a breaking change makes it more
> visible (compile time error), which may help consumers to find incorrect
> handling of these errors.

### v2: remove ErrorHandler and IgnoreNotExist as they are not implemented

Unlike v1, these are not used/implemented in the v2 code, so removing
these.

### DIFF
--- a/v2/errors.go
+++ b/v2/errors.go
@@ -22,16 +22,8 @@ import (
 )
 
 var (
-	ErrInvalidPid               = errors.New("cgroups: pid must be greater than 0")
-	ErrMountPointNotExist       = errors.New("cgroups: cgroup mountpoint does not exist")
-	ErrInvalidFormat            = errors.New("cgroups: parsing file with invalid format failed")
-	ErrFreezerNotSupported      = errors.New("cgroups: freezer cgroup (v2) not supported on this system")
-	ErrMemoryNotSupported       = errors.New("cgroups: memory cgroup (v2) not supported on this system")
-	ErrPidsNotSupported         = errors.New("cgroups: pids cgroup (v2) not supported on this system")
-	ErrCPUNotSupported          = errors.New("cgroups: cpu cgroup (v2) not supported on this system")
-	ErrCgroupDeleted            = errors.New("cgroups: cgroup deleted")
-	ErrNoCgroupMountDestination = errors.New("cgroups: cannot find cgroup mount destination")
-	ErrInvalidGroupPath         = errors.New("cgroups: invalid group path")
+	ErrInvalidFormat    = errors.New("cgroups: parsing file with invalid format failed")
+	ErrInvalidGroupPath = errors.New("cgroups: invalid group path")
 )
 
 // ErrorHandler is a function that handles and acts on errors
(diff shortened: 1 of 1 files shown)


thaJeztah:
As mentioned, this is a **breaking change**, so would require a major version update (if we follow SemVer).

That said, I _think_ the only consumer of the "faulty errors" may be containerd, so it could be considered a "bug" for these to exist (and because of that, "intentional" to introduce a compile time error).

If this is a concern, we can consider _deprecating_ these, and add a warning in the deprecation note that they SHOULD NOT be used.


/cc @AkihiroSuda @crosbymichael @estesp @dims (happy to hear your thoughts)

dims:
I am happy to yank this out @thaJeztah 

- https://cs.k8s.io/?q=github.com%2Fcontainerd%2Fcgroups%2Fv2&i=nope&files=&excludeFiles=&repos=
- https://grep.app/search?q=%22github.com/containerd/cgroups/v2%22&regexp=true
- https://sourcegraph.com/search?q=context:global+%22github.com/containerd/cgroups/v2%22&patternType=literal

LGTM

thanks,
Dims

thaJeztah:
So, we may actually need some handling (_somewhere_) to create if missing (unless cgroup paths act different); see https://github.com/containerd/containerd/pull/5739#issuecomment-880581745

AkihiroSuda (approved):

thaJeztah:
@AkihiroSuda any thoughts on;

> So, we may actually need some handling (_somewhere_) to create if missing (unless cgroup paths act different); see [containerd/containerd#5739 (comment)](https://github.com/containerd/containerd/pull/5739#issuecomment-880581745)

Do you think we need changes here (and _implement_ the `IgnoreNotExist`) or should we change `NewManager()` to be `LoadManager()` -> if not exists -> `NewManager()`?

fuweid (approved):

estesp (approved):
LGTM

estesp:
I'm OK with it being a breaking change. To me, "supporting" a wrong expectation that you might receive these errors is worse than forcing consumers to recognize that isn't happening in the codebase.

thaJeztah:
Yes, I'm fine with that; my only concern was (see above); did we miss something, and should it actually be added after all? If so, then we might as well do so at the same time (and remove the other remaining ones)

dmcgowan (approved):

samuelkarp (approved):
I'm OK with this.  I don't think we've applied SemVer to the client library (or to any of the internal interfaces).  This doesn't change anything about the gRPC API and it doesn't change any of the actual behavior of containerd.

mxpv (approved):

PR #200, opened 2021-07-15
cgroup v1: implement AddProc()

thaJeztah:
This implements AddProc() for cgroups v1 to make the API a closer match with the cgroupsv2 API.

PR #202, opened 2021-07-15
v1: reduce duplicated code

thaJeztah:
~Marking as draft, because this depends on / is a follow-up to https://github.com/containerd/cgroups/pull/200~ (merged)

Code for handling Tasks and Processes, except for "cgroup.procs" vs "tasks", and Tasks being a different type.

This patch:

- Makes the Task type an alias for Process (as they're identical otherwise)
- Adds a `processType` type for the `cgroupProcs` and `cgroupTasks` consts. This type is an alias for "string", and mostly for clarity (indicating it's an 'enum').
- Merges the `cgroup.add()` and `cgroup.addTask()` functions, adding a `pType` argument.
- Merges the `cgroup.processes()` and `cgroup.tasks()` functions, adding a `pType` argument.
- Merges the `readPids()` and `readTasksPids()` utilities, adding a `pType` argument.
- Move locking and validation into `cgroup.add()`. All places using `cgroup.add()` were taking a lock and doing this validation, so looks we can move this code into `cgroup.add()` itself.

### DIFF
--- a/cgroup.go
+++ b/cgroup.go
@@ -151,46 +151,20 @@ func (c *cgroup) Subsystems() []Subsystem {
 
 // Add moves the provided process into the new cgroup
 func (c *cgroup) Add(process Process) error {
-	if process.Pid <= 0 {
-		return ErrInvalidPid
-	}
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	if c.err != nil {
-		return c.err
-	}
-	return c.add(process)
+	return c.add(process, cgroupProcs)
 }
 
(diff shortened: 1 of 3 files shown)


sameo (approved):
FWIW, this looks good  to me.

thaJeztah:
Thanks for reviewing!

Perhaps @AkihiroSuda or @fuweid  could have a look as well (making sure I didn't do anything silly 😂)

AkihiroSuda (approved):

fuweid (approved):
LGTM

PR #203, opened 2021-07-23
cgroup: Optionally add process and task to a subsystems subset

sameo:
In some cases, one may want to add processes and task only to certain
subsystems in a cgroup. For example, when a cgroup is created to only
limit a given subsystem usage, but leaves all other subsystems
unconstrained, there may be a need to move a process to only the
contrained cgroup subsystem while leaving it in another cgroup for
other subsystems.

This commit makes the Control interface process and task addition
functions variadic, in order to pass an optional list of subsystems that
behave as s subsystems filter.

Signed-off-by: Samuel Ortiz <samuel.e.ortiz@protonmail.com>


fuweid:
Need to rebase

sameo:
@fuweid The PR is rebased now.

mxpv (approved):

kzys (approved):

cgroup.go
```diff
@@ -149,22 +149,40 @@ func (c *cgroup) Subsystems() []Subsystem {
...
+		}
+	}
+
+	return filteredSubsystems
+}
+
 // Add moves the provided process into the new cgroup
-func (c *cgroup) Add(process Process) error {
-	return c.add(process, cgroupProcs)
+func (c *cgroup) Add(process Process, subsystems ...Name) error {
```
mikebrow, line 171: Can we update the inline function descriptions too..

control.go
```diff
@@ -54,12 +54,22 @@ type Task = Process
...
+	// the process is added to all the cgroup subsystems. When giving Add a list of
+	// subsystem names, the process is only added to those subsystems, provided that
+	// they are active in the targeted cgroup.
+	Add(Process, ...Name) error
+	// AddProc adds the process with the given id to the cgroup (cgroup.procs).
+	// Without additional arguments, the process with the given id is added to all
+	// the cgroup subsystems. When giving AddProc a list of subsystem names, the process
+	// id is only added to those subsystems, provided that they are active in the targeted
+	// cgroup.
+	AddProc(pid uint64, subsystems ...Name) error
```
mikebrow, line 67: declaration isn't consistent with the other two.. as it includes subsystems

cgroup_test.go
```diff
@@ -101,6 +101,50 @@ func TestAdd(t *testing.T) {
 	}
 }
 
+func TestAddFilteredSubsystems(t *testing.T) {
```
mikebrow, line 104: let's also test against nil filteredSubstems..

mikebrow (commented):
see comments..

Does Update need to be have filters too?

Issue #205, opened 2021-09-07
cgroups.go: Panic on nil interface


linrl3:
https://github.com/containerd/cgroups/blob/main/cgroup.go#L330

If we pass a subsystem that doesn't exist into `func (c *cgroup) Processes(subsystem Name, recursive bool)`, we would get a panic: interface conversion: interface is nil, not cgroups.pather. 

Could I make a patch here to check nil interface before we convert it? Just return error instead of panic.

fuweid:
I think the line 326 can guarantee that subsystem should exist.

linrl3:
@fuweid My code `main.go` was as below. 
```
package main

import (
	"fmt"

	"github.com/containerd/cgroups"
)

var (
	cgroupPath = "/my-test"
	group      = "memory"
)

func main() {
	checkCgroupProcess(cgroupPath, group)
}
func checkCgroupProcess(name string, group string) {
	fmt.Printf("Checking cgroup [%s] process in [%s]\n ", name, group)
	control, err := cgroups.Load(cgroups.V1, cgroups.StaticPath(name))
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(control.Processes(cgroups.Name(group), true))
}
```

Then you can run and got the panic:
```
mkdir /sys/fs/cgroup/cpu/my-test
go run main.go
```

fuweid:
Hmm. The StaticPath doesn't check the path. We can fix with returning ErrControllerNotActive if it is nil. Could you mind to file pr to fix it? Thanks

linrl3:
Sure. Will do it in a few days 👌

linrl3:
Hi @fuweid,

Please review the PR https://github.com/containerd/cgroups/pull/206. Thank you!

fuweid:
close by #207

PR #207, opened 2021-09-08
cgroup.go: avoid panic on nil interface

linrl3:
Hi, this PR's related issue is #205, which is a potential nil interface panic case. 

In https://github.com/containerd/cgroups/blob/main/cgroup.go#L330, if we pass a subsystem that doesn't exist into func (c *cgroup) Processes(subsystem Name, recursive bool), we would get a panic: interface conversion: interface is nil, not cgroups.pather.

So we'd check nil interface and return error if it's nil to avoid panic.

### DIFF
--- a/cgroup.go
+++ b/cgroup.go
@@ -327,6 +327,9 @@ func (c *cgroup) processes(subsystem Name, recursive bool, pType procType) ([]Pr
 	if err != nil {
 		return nil, err
 	}
+	if s == nil {
+		return nil, fmt.Errorf("cgroups: %s doesn't exist in %s subsystem", sp, subsystem)
+	}
 	path := s.(pather).Path(sp)
 	var processes []Process
 	err = filepath.Walk(path, func(p string, info os.FileInfo, err error) error {


cgroup.go
```diff
@@ -323,6 +323,9 @@ func (c *cgroup) Tasks(subsystem Name, recursive bool) ([]Task, error) {
 
 func (c *cgroup) processes(subsystem Name, recursive bool, pType procType) ([]Process, error) {
 	s := c.getSubsystem(subsystem)
+	if s == nil {
+		return nil, ErrControllerNotActive
+	}
 	sp, err := c.path(subsystem)
 	if err != nil {
 		return nil, err
```
fuweid, line 328: I was rethinking that error message. I think the following code might be more clear for the error

After Line 329,

if s == nil {
     return nil, errors.Errorf("no such cgroup %s in subsystem %s", sp, subsystem)
}

WDYT?

linrl3: Could we define a new Error with this?

fuweid: os.ErrNotExist for errors.Wrapf ?

fuweid: I think os.ErrNotExit is fine 😃

linrl3: I was thinking something like 
```
var ErrCgroupNotInSubsystem = errors.New("cgroups: cgroup not in subsystem")
```

fuweid: I was thinking to use templates to generate error by subsystem which make message more clear.

fuweid: > I was thinking something like 
> 
> ```
> 
> var ErrCgroupNotInSubsystem = errors.New("cgroups: cgroup not in subsystem")
> 
> ```

The error doesn't contain the which subsystem and which cgroup. That is My concern

linrl3: Yes, we should make it clear

linrl3: Could we use `fmt.Errorf("cgroups: %s doesn't exist in %s subsystem", sp, subsystem)` ?

fuweid: > Could we use `fmt.Errorf("cgroups: %s doesn't exist in %s subsystem", sp, subsystem)` ?

sgtm

linrl3: A new commit had been pushed.

fuweid (approved):
LGTM

AkihiroSuda (approved):

edboffical:
I'm having the same problem, why not release a new version

PR #208, opened 2021-09-25
replace pkg/errors from vendor

zouyee:
Signed-off-by: Zou Nengren <zouyee1989@gmail.com>


coverage.txt
```diff
@@ -0,0 +1 @@
+mode: atomic
```
AkihiroSuda, line 1: Should be a separate commit

AkihiroSuda (approved):

PR #212, opened 2021-12-03
v2: Fix inotify fd leak when cgroup is deleted

jepio:
This resolves the leak of inotify fd by additionally subscribing to `cgroup.events` notifications, so that we can handle container exits and cgroup deletion. This is best tested in the context of k8s, the same manifest as https://github.com/containerd/containerd/pull/6323 can be used  - increase the replica count and watch the inotify instances get exhausted.

See issue: https://github.com/containerd/containerd/issues/5670


v2/manager.go
```diff
@@ -572,6 +589,13 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
 		syscall.Close(fd)
 		return 0, 0, fmt.Errorf("failed to add inotify watch for %q", fpath)
 	}
+	// monitor to detect process exit/cgroup deletion
+	evpath := filepath.Join(c.path, "cgroup.events")
+	wd_cg, err := syscall.InotifyAddWatch(fd, evpath, unix.IN_MODIFY)
```
AkihiroSuda, line 594: - err seems unchecked
- nit: underscores are usually avoided for variable names in Go

jepio: Thanks, didn't know the thing about underscores. Added error checking in both `InotifyAddWatch` calls, and replaced `wd_cg` with `_` since it is discarded anyway.

v2/manager.go
```diff
@@ -568,9 +585,16 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
...
+		syscall.Close(fd)
+		return 0, 0, fmt.Errorf("failed to add inotify watch for %q: %s", fpath, err)
+	}
+	// monitor to detect process exit/cgroup deletion
+	evpath := filepath.Join(c.path, "cgroup.events")
+	_, err = syscall.InotifyAddWatch(fd, evpath, unix.IN_MODIFY)
+	if err != nil {
 		syscall.Close(fd)
-		return 0, 0, fmt.Errorf("failed to add inotify watch for %q", fpath)
+		return 0, 0, fmt.Errorf("failed to add inotify watch for %q: %s", evpath, err)
```
AkihiroSuda, line 597: %w is preferred for wrapping err

v2/manager.go
```diff
@@ -568,9 +585,16 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
 		return 0, 0, errors.New("failed to create inotify fd")
 	}
 	wd, err := syscall.InotifyAddWatch(fd, fpath, unix.IN_MODIFY)
-	if wd < 0 {
+	if err != nil {
+		syscall.Close(fd)
+		return 0, 0, fmt.Errorf("failed to add inotify watch for %q: %s", fpath, err)
```
AkihiroSuda, line 590: %w is preferred for wrapping err

v2/manager.go
jepio: Switched to %w now.

v2/manager.go
jepio: Switched to %w now.

AkihiroSuda (approved):

jepio:
Hi @AkihiroSuda, can this be merged? Does this repository need to be tagged so that it can be updated in containerd/containerd?

AkihiroSuda:
> Hi @AkihiroSuda, can this be merged? 

Needs another LGTM from the committers.

> Does this repository need to be tagged so that it can be updated in containerd/containerd?

Yes, we will probably create a tag right after merging this.

ktock (approved):
LGTM

v2/manager.go
```diff
@@ -568,9 +585,16 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
 		return 0, 0, errors.New("failed to create inotify fd")
 	}
 	wd, err := syscall.InotifyAddWatch(fd, fpath, unix.IN_MODIFY)
-	if wd < 0 {
+	if err != nil {
+		syscall.Close(fd)
```
fuweid, line 589: is it possible to use defer for cleanup?

v2/manager.go
```diff
@@ -581,13 +605,12 @@ func (c *Manager) EventChan() (<-chan Event, <-chan error) {
 	errCh := make(chan error)
```
fuweid, line 581: IMO, one buffer size should be good here. If no one reads the channel, the waitForEvents will be blocked in some condition.

v2/manager.go
```diff
@@ -581,13 +605,12 @@ func (c *Manager) EventChan() (<-chan Event, <-chan error) {
...
-	return ec, nil
+	return ec, errCh
 }
 
 func (c *Manager) waitForEvents(ec chan<- Event, errCh chan<- error) {
-	fd, wd, err := c.MemoryEventFD()
+	fd, _, err := c.MemoryEventFD()
 
-	defer syscall.InotifyRmWatch(fd, wd)
 	defer syscall.Close(fd)
```
fuweid, line 591: Could we call the defer after error check?

v2/manager.go
```diff
@@ -643,7 +666,17 @@ func (c *Manager) waitForEvents(ec chan<- Event, errCh chan<- error) {
 				}
 				ec <- e
 			} else {
-				errCh <- err
+				// When cgroup is deleted read may return -ENODEV instead of -ENOENT from open
+				_, statErr := os.Lstat(filepath.Join(c.path, "memory.events"))
```
fuweid, line 670: Maybe we can use 

If !os.IsNotExist() {
    errCh <- err
}

And one question here: why not just call the close channel in defer?

jepio: I rewrote it as
```
if  !os.IsNotExist(statErr) {
   errCh <- err
}
```
And added the defer channel close (but its the first line in the function so that it happens with every return.

fuweid (approved):
LGTM

invidian (commented):
Should there be some tests for this behavior change so we make sure it does not regress?

jbehling:
Any word on when we can get this merged? This is blocking us from rolling out cgroups v2

jepio:
@invidian you're welcome to add tests, I'm currently working on other things and don't have the time.

Is there any one else that wants to review this? I was hoping to get this into container 1.6.0 - https://github.com/containerd/containerd/issues/5670 is marked as a priority 1 bug.

invidian:
>@invidian you're welcome to add tests, I'm currently working on other things and don't have the time.

I can have a look, but no promises and I guess it shouldn't be blocking this PR.

v2/manager.go
```diff
@@ -568,32 +585,38 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
 		return 0, 0, errors.New("failed to create inotify fd")
 	}
 	wd, err := syscall.InotifyAddWatch(fd, fpath, unix.IN_MODIFY)
-	if wd < 0 {
+	if err != nil {
 		syscall.Close(fd)
-		return 0, 0, fmt.Errorf("failed to add inotify watch for %q", fpath)
+		return 0, 0, fmt.Errorf("failed to add inotify watch for %q: %w", fpath, err)
+	}
```
invidian, line 591: This change is unrelated I think and could be done at least in a separate commit. Tests shouldn't be hard either, I think you just need to pass it an empty string (so it has NULL as first byte).

v2/manager.go
```diff
@@ -568,32 +585,38 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
...
+	// monitor to detect process exit/cgroup deletion
+	evpath := filepath.Join(c.path, "cgroup.events")
+	_, err = syscall.InotifyAddWatch(fd, evpath, unix.IN_MODIFY)
```
invidian, line 594: This means number of watches will be doubled for all existing systems, right?

v2/manager.go
```diff
@@ -568,32 +585,38 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
 		return 0, 0, errors.New("failed to create inotify fd")
 	}
 	wd, err := syscall.InotifyAddWatch(fd, fpath, unix.IN_MODIFY)
-	if wd < 0 {
+	if err != nil {
 		syscall.Close(fd)
```
invidian, line 589: I wonder if there is a way we could handle the possible errors coming from this `Close()` call.

v2/manager.go
```diff
@@ -643,7 +666,14 @@ func (c *Manager) waitForEvents(ec chan<- Event, errCh chan<- error) {
 				}
 				ec <- e
 			} else {
-				errCh <- err
+				// When cgroup is deleted read may return -ENODEV instead of -ENOENT from open
```
invidian, line 669:
```suggestion
				// When cgroup is deleted read may return -ENODEV instead of -ENOENT from open.
```

v2/manager.go
```diff
@@ -643,7 +666,14 @@ func (c *Manager) waitForEvents(ec chan<- Event, errCh chan<- error) {
...
+				_, statErr := os.Lstat(filepath.Join(c.path, "memory.events"))
+				if !os.IsNotExist(statErr) {
```
invidian, line 671:
```suggestion
				if _, statErr := os.Lstat(filepath.Join(c.path, "memory.events")); !os.IsNotExist(statErr) {
```

v2/manager.go
```diff
@@ -560,6 +560,23 @@ func (c *Manager) freeze(path string, state State) error {
 	}
 }
 
+func (c *Manager) isCgroupEmpty() bool {
+	// In case of any error we return true so that we exit and don't leak resources
+	out := make(map[string]interface{})
+	err := readKVStatsFile(c.path, "cgroup.events", out)
+	if err != nil {
```
invidian, line 567:
```suggestion
	if err := readKVStatsFile(c.path, "cgroup.events", out); err != nil {
```

v2/manager.go
```diff
@@ -560,6 +560,23 @@ func (c *Manager) freeze(path string, state State) error {
...
+		return true
+	}
+	if v, ok := out["populated"]; ok {
+		populated, ok := v.(uint64)
+		if !ok {
+			return true
+		}
+		return populated == 0
```
invidian, line 575: This changes the behavior of `EventChan()` if I read code correctly. In the following scenario:
1. Create cgroup Z.
2. Move process A to cgroup Z.
3. Terminate process A.
4. Move process B to cgroup Z.

After step 3. the watching loop will exit and be cleaned up, which is currently not the case. Is that desired?

v2/manager.go
```diff
@@ -568,32 +585,38 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
...
-	return ec, nil
+	return ec, errCh
 }
 
 func (c *Manager) waitForEvents(ec chan<- Event, errCh chan<- error) {
-	fd, wd, err := c.MemoryEventFD()
-
-	defer syscall.InotifyRmWatch(fd, wd)
-	defer syscall.Close(fd)
+	defer close(errCh)
```
invidian, line 612: This may potentially be a breaking change. If users of `EventChan()` unconditionally close the error channel, it will make them panic with `close of closed channel` error.

v2/manager.go
```diff
@@ -643,7 +666,14 @@ func (c *Manager) waitForEvents(ec chan<- Event, errCh chan<- error) {
 				}
 				ec <- e
 			} else {
-				errCh <- err
```
invidian, line 646: Likely out of scope, but it seems flipping `if err := readKVStatsFile(c.path, "memory.events", out); err == nil {` logic would allow to save bunch of indentation and make code much more readable and with happy path left-aligned.

v2/manager.go
```diff
@@ -568,32 +585,38 @@ func (c *Manager) MemoryEventFD() (int, uint32, error) {
...
-	if wd < 0 {
+	if err != nil {
 		syscall.Close(fd)
-		return 0, 0, fmt.Errorf("failed to add inotify watch for %q", fpath)
+		return 0, 0, fmt.Errorf("failed to add inotify watch for %q: %w", fpath, err)
+	}
+	// monitor to detect process exit/cgroup deletion
+	evpath := filepath.Join(c.path, "cgroup.events")
+	_, err = syscall.InotifyAddWatch(fd, evpath, unix.IN_MODIFY)
+	if err != nil {
```
invidian, line 595:
```suggestion
	if _, err := syscall.InotifyAddWatch(fd, evpath, unix.IN_MODIFY); err != nil {
```

invidian (commented):
Quick draft of test which could be written, though it relies on `go.uber.org/goleak` for detecting leaking goroutines, maybe that could be improved. This currently fails on `main` branch , but passes on this PR. But this PR modifies much more so perhaps more tests should be added.

While reading the code and writing tests for it I also spotted some potentially breaking changes, so mentioned them in the comments.

```go
package v2

import (
  "fmt"
  "io/fs"
  "os"
  "path/filepath"
  "strings"
  "testing"
  "time"

  "go.uber.org/goleak"
)

func Test_EventWatch_stops_watching_memory_events_when_cgroup_is_removed(t *testing.T) {
  skipWhenNotRunAsRoot(t)
  checkCgroupMode(t)

  group := "/cpu-test-cg"
  groupPath := fmt.Sprintf("%s-%d", group, os.Getpid())

  c, err := NewManager(defaultCgroup2Path, groupPath, &Resources{})
  if err != nil {
    t.Fatal("failed to init new cgroup manager: ", err)
  }

  evCh, errCh := c.EventChan()

  data := `low 0
high 0
max 0
oom 0
oom_kill 0
`

  ticker := time.NewTicker(100 * time.Millisecond)
  done := make(chan bool)

  go func() {
    for {
      select {
      case <-done:
        return
      case <-ticker.C:
        err := os.WriteFile(filepath.Join(defaultCgroup2Path, groupPath, "memory.events"), []byte(data), 0o444)
        if !strings.HasSuffix(err.Error(), fs.ErrInvalid.Error()) {
          t.Logf("Unexpected error writing event to cgroup: %v", err)
          t.Fail()
        }
      }
    }
  }()

  select {
  case ev := <-evCh:
    t.Logf("event received: %v", ev)
  case err := <-errCh:
    if err != nil {
      t.Fatalf("Received error event: %v", err)
    }
  }

  t.Cleanup(func() {
    if err := os.Remove(filepath.Join(defaultCgroup2Path, groupPath)); err != nil {
      t.Fatalf("Removing cgroup path failed: %v", err)
    }

    ticker.Stop()
    done <- true

    goleak.VerifyNone(t)
  })
}
```

v2/manager.go
invidian: Ah, good to know the difference between instances and watches! I was thinking that there might be some correlation between watches and max pods on node configurable or something. Seems OK, feel free to mark as resolved :+1:

v2/manager.go
invidian: Make sense, thanks for explanation!

v2/manager.go
```diff
@@ -604,46 +625,51 @@ func (c *Manager) waitForEvents(ec chan<- Event, errCh chan<- error) {
...
-					if !ok {
-						errCh <- fmt.Errorf("cannot convert high to uint64: %+v", v)
-						return
-					}
+			if err := readKVStatsFile(c.path, "memory.events", out); err != nil {
+				// When cgroup is deleted read may return -ENODEV instead of -ENOENT from open.
+				if _, statErr := os.Lstat(filepath.Join(c.path, "memory.events")); !os.IsNotExist(statErr) {
+					errCh <- err
+				}
+				return
```
invidian, line 633: Also not directly related, but moving all the code to a separate function returning an error would also make code much more readable, as then you can use more convenient `return err` instead of send to channel + naked return.

invidian (commented):
Thanks for addressing my comments @jepio!

dmcgowan:
The `vendor` directory got pulled into one of the commits.

v2/manager_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"go.uber.org/goleak"
+)
+
+func TestEventChanCleanupOnCgroupRemoval(t *testing.T) {
+	checkCgroupMode(t)
+
+	cmd := exec.Command("cat")
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		t.Fatal("failed to create cat process: ", err)
```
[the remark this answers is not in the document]
jepio: Applied that to the test case, but not to `v2/manager.go` because the rest of the code uses lower-case as a convention.

v2/manager.go
jepio: Moved the code to a function (`parseMemoryEvents`).

v2/manager_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"go.uber.org/goleak"
+)
+
+func TestEventChanCleanupOnCgroupRemoval(t *testing.T) {
```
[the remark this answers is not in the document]
jepio: There was a race in the test case, simplified it the way we agreed offline and now this should not happen. Try the updated test case.

jepio:
@dmcgowan got rid of the vendor directory again, I believe that should cover everything.

Vagrantfile
```diff
@@ -15,7 +15,7 @@ Vagrant.configure("2") do |config|
   config.vm.provision "shell", inline: <<-SHELL
     set -eux -o pipefail
     # configuration
-    GO_VERSION="1.15"
+    GO_VERSION="1.16"
```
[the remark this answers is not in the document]
jepio: Reverted to GO=1.15, and os.ErrInvalid is not used any longer.

dmcgowan (approved):

v2/manager_test.go
```diff
@@ -0,0 +1,79 @@
...
+				t.Fatalf("Unexpected error on error channel: %v", err)
+			}
+			done = true
+		case <-time.After(5 * time.Second):
+			t.Fatal("Timed out")
+		}
+		if done {
+			break
+		}
+	}
```
dcantah, line 77: nit: Could be 
```go 
for done {
	select {
		case <-evCh:
		case err := <-errCh:
			if err != nil {
				t.Fatalf("Unexpected error on error channel: %v", err)
			}
                        done = true
		case <-time.After(5 * time.Second):
			t.Fatal("Timed out")
	}
}

dcantah (approved):
LGTM, minor nitpick

v2/manager_test.go
jepio: Oh, yes. That's what i was looking for. Still getting used to the lack of "while".

estesp (approved):
LGTM

PR #220, opened 2022-02-17
Use google.golang.org/protobuf instead of github.com/gogo/protobuf

kzys:
We'd like to migrate off from gogo/protobuf. The package is looking
for new ownership since 2020.

See https://github.com/containerd/containerd/issues/6564 for the detail.


kzys:
This PR is ready, but it would break consumers that use cgroups from protos, such as https://github.com/Microsoft/hcsshim. We need to cut a new branch for backporting bug fixes.

estesp:
So, we can do the same thing here we do in containerd (have a `/release/1.0.x` branch anchored at the recent v1.0.3 tag that cuts all future 1.0.x releases) but I think this would be the first subproject to get that treatment. Seems like it is somewhat inevitable as we will have overlap of containerd releases that use both old and new protobuf. Does that seem right?

kzys:
@estesp Yes. `containerd/cgroups` and `containerd/ttrpc` would have a new branch (ttrpc is being discussed in https://github.com/containerd/ttrpc/issues/97).

Makefile
```diff
@@ -22,3 +22,5 @@ cgutil:
 
 proto:
 	protobuild --quiet ${PACKAGES}
+	go-fix-acronym -w -a Cpu -a Rss -a Tcp \
```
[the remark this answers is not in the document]
kzys: And it is merged!

dmcgowan (approved):

thaJeztah:
I see there's a couple of fixes in main that have been merged (https://github.com/containerd/cgroups/compare/v1.0.3... f8328fdc061d6fb757b370c19ad8f0d05a281a41); but not yet released. Given that this change could be (potentially?) risky, I'm wondering;

- should we do a new tag with the other changes first (before merging this?)
- what version should we use for the changes of this PR? (should we make it `v1.1.x` is it "risky" enough to warrant `v2.x.x`)? Mostly considering there to keep the option of doing patch/fix release for the existing `v1.(0).x` releases if there's a need

thaJeztah:
> I'm inclined to have 1.1.x. It should not impact most of clients.

Given that v1.1.0 would be considered a minor (feature only, backward compatible) update, there is a chance that repositories automatically update to that version; should we do a test in some repositories (e.g. a draft PR that updates the containerd 1.5 branch to use this version) to check if it doesn't break?

thaJeztah:
> Yes. I was doing [containerd/containerd#6563](https://github.com/containerd/containerd/pull/6563) but can't recall the reason to upgrade hcsshim. Let me check.

Ah! Think I missed that one. I hope that one works (also let me know if you want me to try this branch on other repositories). 

If it _doesn't_ work, I think the only "good" option would be to make it a V2, taking into account that we have circular dependencies in various repositories, which makes updating to a non-compatible version more challenging (having a v2 means they can potentially co-exist during the transition).

kzys:
Regarding hcsshim, the shim is embedding this cgroup package's message is its own message. If one of them is generated by gogo (or vice versa, probably), a protobuf serializer couldn't handle another.

https://github.com/microsoft/hcsshim/blob/7fa8bda4e6ba503caf0d53d0a4ee99b9a64ceed8/cmd/containerd-shim-runhcs-v1/stats/stats.proto#L11-L17

kzys:
Let me convert that to draft for a moment. I'd like to sync-up with hcsshim folks first to see how to handle this upgrade in there.

kzys:
Working on https://github.com/microsoft/hcsshim/pull/1356, which results https://github.com/containerd/protobuild/pull/52.

Makefile
```diff
@@ -22,3 +22,4 @@ cgutil:
 
 proto:
 	protobuild --quiet ${PACKAGES}
+	gopatch -p rename.gopatch stats/v1/ v2/stats/
```
kzys, line 25: Let me update this line.

AkihiroSuda:
What’s current status of this?

kzys:
seems hcsshim can have the gogo-less cgroups without upgrading containerd. So this PR is be good to go.

https://github.com/microsoft/hcsshim/pull/1356

dmcgowan (approved):

.github/workflows/ci.yml
```diff
@@ -112,3 +112,42 @@ jobs:
...
+
+  proto:
+    name: Compare auto-generated Go files
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.17.x'
```
AkihiroSuda, line 124: Any reason not to use 1.19 ?

.github/workflows/ci.yml
```diff
@@ -112,3 +112,42 @@ jobs:
...
+
+      - name: Set env
+        shell: bash
+        run: |
+          echo "GOPATH=${{ github.workspace }}" >> $GITHUB_ENV
+          echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
+
+      - name: Checkout cgroups
+        uses: actions/checkout@v2
```
AkihiroSuda, line 133: Can we use v3?

AkihiroSuda (approved):

.github/workflows/ci.yml
kzys: Because this PR is old. Updated.

.github/workflows/ci.yml
kzys: Sure.

estesp (approved):
LGTM

thaJeztah:
🎉 awesome work, @kzys !

AkihiroSuda:
Shall we release v1.1.0?

kzys:
@AkihiroSuda After reading https://semver.org/ again and working on hcsshim, I'm more inclined to have v2.0.0. Thoughts?

thaJeztah:
v2 sounds like a good idea, that would keep the path open to have projects opt-in to the new thing. But it will require the module to be renamed to /v2/

kzys:
Ugh. We will have github.com/containerd/cgroups/v2/v2/stats by doing so, but we would reach the state eventually anyway...

thaJeztah:
Given that it's v2, it would allow us to rename that package to something else 🤔 

(naming is hard though!)

Perhaps we could do the reverse, and rename the v1 cgroups? (as v2 basically is "default" now, so "v1" will more and more become the outlier)

stats/v1/metrics.pb.go
```diff
@@ -315,5811 +488,1472 @@ type MemoryStat struct {
...
-	proto.RegisterType((*Throttle)(nil), "io.containerd.cgroups.v1.Throttle")
-	proto.RegisterType((*MemoryStat)(nil), "io.containerd.cgroups.v1.MemoryStat")
-	proto.RegisterType((*MemoryEntry)(nil), "io.containerd.cgroups.v1.MemoryEntry")
-	proto.RegisterType((*MemoryOomControl)(nil), "io.containerd.cgroups.v1.MemoryOomControl")
-	proto.RegisterType((*BlkIOStat)(nil), "io.containerd.cgroups.v1.BlkIOStat")
-	proto.RegisterType((*BlkIOEntry)(nil), "io.containerd.cgroups.v1.BlkIOEntry")
-	proto.RegisterType((*RdmaStat)(nil), "io.containerd.cgroups.v1.RdmaStat")
-	proto.RegisterType((*RdmaEntry)(nil), "io.containerd.cgroups.v1.RdmaEntry")
-	proto.RegisterType((*NetworkStat)(nil), "io.containerd.cgroups.v1.NetworkStat")
-	proto.RegisterType((*CgroupStats)(nil), "io.containerd.cgroups.v1.CgroupStats")
```
AkihiroSuda, line 719: Where has this gone?

PR #226, opened 2022-03-21
make cmd/ a separate module (as it's only for testing)

thaJeztah:
To my knowledge, the `cmd/cgctl` utility is only for testing (also see https://github.com/containerd/cgroups/commit/7bf30d7bf13852d1f2b493817489d37170d2b8a4), and its dependencies should therefore not be needed as dependencies for the main module.

This patch adds a `go.mod` in the `cmd/` directory to make it it's own module. A replace rule is included in that `go.mod` to make it use the local version of github.com/containerd/cgroups when building.

### DIFF
--- a/cmd/go.mod
+++ b/cmd/go.mod
@@ -1,0 +1,11 @@
+module github.com/containerd/cgroups/cmd
+
+go 1.16
+
+replace github.com/containerd/cgroups => ../
+
+require (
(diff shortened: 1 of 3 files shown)


kzys (approved):
LGTM!

dmcgowan (approved):

PR #227, opened 2022-03-23
Do not override devices in cgv2 because of default settings by runc

utam0k:
Update of devices in cgroup v2 will override runc settings. This is not an addition, but a complete overwrite. Therefore, it will overwrite the default settings. I'd like to fix this if I will have the time but once I tried to disable it. Let me know what the community thinks.

https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#default-devices
https://github.com/opencontainers/runc/blob/2436322fef995f60194554c5ce2947df2e18661d/libcontainer/specconv/spec_linux.go#L188

### DIFF
--- a/v2/manager.go
+++ b/v2/manager.go
@@ -241,6 +241,10 @@ func (c *Manager) Controllers() ([]string, error) {
 }
 
 func (c *Manager) Update(resources *Resources) error {
+	// skip updating devices because the default devices setting is by runc.
+	if len(resources.Devices) > 0 {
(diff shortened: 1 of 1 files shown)


kzys (commented):
Hmm, not so sure. The method itself shouldn't assume runc or any OCI runtimes.

kzys:
@utam0k While the biggest user of this package is containerd, the package itself doesn't assume runc or any other runtimes. Is the current behavior blocking some use cases?

PR #228, opened 2022-04-11
Check that cgroup is empty before deleting

jseba:
The kernel will block an attempt to rmdir a cgroup path that still has running processes in it. Since the removal code uses the standard `os.RemoveAll` function, the Go runtime will helpfully fall back to an `rm -rf`-like removal algorithm if the cheap `unlink`/`rmdir` attempts fail.

Since cgroup files cannot be removed, the caller ends up with an unhelpful "unlinkat /sys/fs/cgroup/.../cgroup.events: operation not permitted" error message that doesn't actually give any actionable information because the original error was swallowed by the runtime.

This changes the `Delete` functions to detect cgroups with still running processes and return an error indicating that removal cannot be done.

Signed-off-by: Josh Seba <sebajosh@outlook.com>

### DIFF
--- a/cgroup.go
+++ b/cgroup.go
@@ -229,7 +229,7 @@ func (c *cgroup) Delete() error {
 	var errs []string
 	for _, s := range c.subsystems {
 		// kernel prevents cgroups with running process from being removed, check the tree is empty
-		procs, err := c.Processes(s.Name(), true)
+		procs, err := c.processes(s.Name(), true, cgroupProcs)
 		if err != nil {
 			return err
 		}
--- a/cgroup.go
+++ b/cgroup.go
@@ -234,7 +234,7 @@ func (c *cgroup) Delete() error {
 			return err
 		}
 		if len(procs) > 0 {
-			errs = append(errs, string(s.Name()))
+			errs = append(errs, fmt.Sprintf("%s (contains running processes)", string(s.Name())))
 			continue
 		}
 		if d, ok := s.(deleter); ok {
--- a/cgroup.go
+++ b/cgroup.go
@@ -84,7 +84,7 @@ func Load(hierarchy Hierarchy, path Path, opts ...InitOpts) (Cgroup, error) {
 	for _, s := range pathers(subsystems) {
 		p, err := path(s.Name())
 		if err != nil {
-			if  errors.Is(err, os.ErrNotExist) {
+			if errors.Is(err, os.ErrNotExist) {
 				return nil, ErrCgroupDeleted
 			}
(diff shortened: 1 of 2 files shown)


v2/manager.go
```diff
@@ -337,6 +337,14 @@ func (c *Manager) AddProc(pid uint64) error {
 }
 
 func (c *Manager) Delete() error {
+	// kernel prevents cgroups with running process from being removed, check the tree is empty
+	processes, err := c.Procs(true)
+	if err != nil {
+		return err
+	}
+	if len(processes) > 0 {
+		return fmt.Errorf("cgroups: unable to remove path %q: still contains running processes", c.path)
```
kzys, line 346: Is it possible to have a similar message on cgroup.go?

jseba: The v1 Delete just accumulates a list of subsystems that couldn't be removed and doesn't report the error. I could make the append more like:

`errs = append(errs, fmt.Sprintf("%s (contains running processes)", string(s.Name()))`

be a solution? Then the resulting error would be akin to

`cgroups: unable to remove paths memory (contains running processes), cpu (contains running processes)`

kzys: sounds good to me!

jseba: Apologies for not getting back to this sooner, I added the "contains running process" reason to the errors list in cgroup.go

jseba:
If someone gets a chance to take a look at this again, I'd appreciate it! It bubbled up to the top of our ticket tracker to remind me to check on it. If there's any blockers let me know!

cgroup.go
```diff
@@ -247,6 +256,7 @@ func (c *cgroup) Delete() error {
 			if err := remove(path); err != nil {
 				errs = append(errs, path)
 			}
+			continue
```
estesp, line 259: is this `continue` necessary? I don't see how it changes loop flow since this is the end of the `for` and will naturally continue through the end of the range of subsystems.

jseba: It's not strictly necessary, no. I added it to because from my interpretation of the structure of the code here, it's sort of like a switch case and only one branch should be valid, so if there was ever a third case (very unlikely since cgroup v1 is now legacy code), this would be correct for that situation.

Looking at it a bit more, I think it would be better if those two `if` blocks were restructured to better represent that intent that only one interface is expected to be matched. What do you think of something more like this, would that be more readable?
```go
switch s.(type) {
case deleter:
    ...
case pather:
   ...
}
```

kzys: Checked @estesp offline and this is not a blocker.

PR #233, opened 2022-06-03
Update README to include usage examples of v2 client

sparrc:
Recently AWS ECS pushed support for cgroups v2 using the cgroupsv2 client library provided here:
https://github.com/aws/amazon-ecs-agent/pull/3161. This PR is based on how I learned to use the client library and hope it can help other devs to do the same.


estesp (approved):
LGTM. Thanks!

kzys (approved):

PR #246, opened 2022-10-17
Don't use ioutil

kzys:
io/ioutil has been deprecated since Go 1.16.


estesp (approved):
LGTM

PR #249, opened 2022-10-24
Upgrade GitHub actions CI workflow

austinvazquez:
Updates the GitHub action CI workflow to use the ubuntu-22.04 image since the ubuntu-18.04 image has been marked deprecated. Also updates the actions/checkout and actions/setup-go packages to the latest version.

Related issue: https://github.com/containerd/project/issues/95

Signed-off-by: Austin Vazquez <macedonv@amazon.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
   #
   project:
     name: Project Checks
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     timeout-minutes: 5
 
     steps:
@@ -23,7 +23,7 @@ jobs:
       # Install Go
       #
       - name: Install Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
           go-version: '1.17.x'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -85,11 +85,7 @@ jobs:
     timeout-minutes: 15
     needs: [project]
 
-    strategy:
-      matrix:
-        os: [ubuntu-18.04, ubuntu-22.04]
```
austinvazquez, line 90: Both 20.04 and 22.04 images use the same Linux kernel version, 5.15.0-1022-azure, so it didn't make sense to test both.

estesp:
@AkihiroSuda do we lose cgroups v1 testing if we move to 22.04? I thought I remembered that 22.04 defaults to cgroups v2 but maybe I'm wrong.

AkihiroSuda:
> @AkihiroSuda do we lose cgroups v1 testing if we move to 22.04? I thought I remembered that 22.04 defaults to cgroups v2 but maybe I'm wrong.

Yes, so we need to use 20.04, not 22.04

AkihiroSuda (changes requested):
https://github.com/containerd/cgroups/pull/249#issuecomment-1293765241

austinvazquez:
Thanks for the clarification @AkihiroSuda. Added a comment to remind the next person once Ubuntu 20.04 goes EOL.

AkihiroSuda (approved):

PR #250, opened 2022-11-04
Retry on EINVAL like runc

kzys:
There is a race condition while adding a brand-new process to cgroups.procs. Like runc, we should retry on EINVAL.

We have seen the issue a lot on containerd's Rocky Linux build.
https://github.com/containerd/containerd/issues/7532

### DIFF
--- a/cgroup.go
+++ b/cgroup.go
@@ -19,11 +19,14 @@ package cgroups
 import (
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
+	"time"
 
 	v1 "github.com/containerd/cgroups/stats/v1"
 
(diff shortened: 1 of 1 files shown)


kzys:
This could be the first issue to backport ;)

cgroup.go
```diff
@@ -193,6 +196,31 @@ func (c *cgroup) AddTask(process Process, subsystems ...Name) error {
...
 
+// writeCgroupsProcs writes to the file, but retries on EINVAL.
+func writeCgroupProcs(path string, content []byte, perm fs.FileMode) error {
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, perm)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	for i := 0; i < 5; i++ {
```
dcantah, line 207: With how this is structured we'd retry even if the error we got wasn't EINVAL, right (just with no delay)? Should we just return the error directly if != EINVAL? The continue shouldn't be needed if we want to keep this how it is

cgroup.go
```diff
@@ -207,7 +235,7 @@ func (c *cgroup) add(process Process, pType procType, subsystems ...Name) error
 		if err != nil {
 			return err
 		}
-		err = retryingWriteFile(
+		err = writeCgroupProcs(
```
dcantah, line 238: Do we still need the EINTR checks that retryingWriteFile does? In 1.15 Go handles this for us and will retry, but it was never back ported to 1.14 afaict where EINTR is more common as well. Not sure if we even use anything in this lib that would block someone on 1.14 from using it

cgroup.go
dcantah: I guess this kind of plays into my [other comment](https://github.com/containerd/cgroups/pull/250/files#r1014478984)

kzys: Oh right. Will fix.

cgroup.go
dcantah: Oh nice, we use os.ReadFile and friends now from 1.16 https://github.com/containerd/cgroups/commit/98dc9d945cc4026ec441e3416a12c4899d45f0b4

kzys: No. We can remove retryingWriteFile() now.

https://github.com/golang/go/issues/38033

dcantah: So truthfully we should be able to get rid of that retryingWriteFile throughout I believe. I can look into that

dcantah: Awesome. Seems like this pkg is 1.16 minimum because we got off of ioutil, and the eintr fix was in 1.15. I'll drop retryingWriteFile from the rest of the project

dcantah (approved):
Thanks! LGTM

estesp:
needs a rebase to pick up @dcantah's change I guess?

kzys:
@estesp Yes. Just rebased.

PR #251, opened 2022-11-04
Rename the package to github.com/containerd/cgroups/v2

kzys:
Migrating off from gogo/protobuf is backward-incompatible for some
consumers that use proto-generated structs such as hcsshim.


v2/manager.go
```diff
@@ -30,7 +30,7 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/containerd/cgroups/v2/stats"
+	"github.com/containerd/cgroups/v2/v2/stats"
```
kzys, line 33: This is ugly, but if we move types from v2/v2 to v2, having cgroups v3 (if there is) is a headache. So I'd stay with this simple-yet-ugly naming.

kzys: Having github.com/containerd/cgroups/v2/cgroup2 (instead of v2/v2) could be an option, by following the filesystem type.

https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html

AkihiroSuda: Can we just rename that to “unified” ?

AkihiroSuda: The cgroup v1 impl can be named “legacy”

.github/workflows/ci.yml
```diff
@@ -144,7 +147,7 @@ jobs:
       - name: Install proto-related tools for Go
         run: |
           go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28
-          go install github.com/containerd/protobuild@v0.2.0
+          go install github.com/containerd/protobuild@14832ccc41429f5c4f81028e5af08aa233a219cf
```
AkihiroSuda, line 150: Can we create a new release of protobuild?

v2/manager.go
kzys: `unified` may work, but `legacy` is not future-proof. Would cgroup v2 be considered as legacy eventually?

.github/workflows/ci.yml
kzys: Sure

kzys: I'd like to include https://github.com/containerd/protobuild/pull/56 in 0.3.0.

v2/manager.go
kzys: Hmm, but unified.NewManager doesn't look better than cgroups.NewManager.

kzys:
I'm open to move cgroup v1 files to cgroup1 directory, but I'd like to do that in a different PR since this one already touches 47 files.

dmcgowan:
> I'm open to move cgroup v1 files to cgroup1 directory, but I'd like to do that in a different PR since this one already touches 47 files.

Does that change need to happen in this PR, this PR looks good. We can make the naming consistent in another PR since they will be separate commits anyway.

kzys:
> Does that change need to happen in this PR

I don't think so.

.github/workflows/ci.yml
kzys: Done.

v2/manager.go
kzys: I'm settled to use cgroup2 now.

AkihiroSuda:
Can we now drop "v1" from github.com/containerd/cgroups/v2/stats/v1 ?

kzys:
> Can we now drop "v1" from github.com/containerd/cgroups/v2/stats/v1 ?

Yes. How about doing the following in a follow-up PR?

- Move cgroups/v2/*.go to cgroups/v2/cgroup1
- Move cgroups/v2/stats/v1 to cgroups/v2/cgroup1/stats

Then we have the most under cgroups/v2/cgroup1 and cgroups/v2/cgroup2.

AkihiroSuda (approved):

AkihiroSuda:
> How about doing the following in a follow-up PR?

SGTM

I thought it can be just done in this PR with a separate commit but either is fine to me

PR #255, opened 2022-11-08
Add cgroup1/ to make cgroup1 and cgroup2 symmmetric

kzys:


cgroup1/state.go
```diff
@@ -14,7 +14,7 @@
    limitations under the License.
 */
 
-package cgroups
+package cgroup1
 
 // State is a type that represents the state of the current cgroup
 type State string
```
kzys, line 20: Does it make sense to share types like State between cgroup1 and cgroup2?

AkihiroSuda: Not necessary to share

AkihiroSuda (approved):

PR #259, opened 2022-11-11
Make cgroup1.Load and cgroup2.Load closer

kzys:
- cgroup1.Load's first parameter could be moved to functional options.
- cgroup2.LoadManager's first parameter could be moved too and Manager suffix doesn't make much sense.

Both changes make cgroup1.Load and cgroup2.Load much closer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ uses the v1 implementation of cgroups.
 
 ```go
 shares := uint64(100)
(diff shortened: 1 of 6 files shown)


AkihiroSuda (approved):

AkihiroSuda:
README.md needs to be updated too for the new API

PR #263, opened 2022-12-31
cgroup2: Add Kill method to manager

dcantah:
This adds in support for killing all of the processes in a cgroup. In 5.14+ this is very simple, a cgroup.kill file exists that all you need to do is write "1" to https://lwn.net/Articles/855924/. On kernels prior, or if the file doesn't exist to be more pedantic to account for potential backports, I've taken the approach runc currently uses which is a manual process of freezing the cgroup -> sending a signal to all of the processes -> thawing the cgroup.

This also adds in a simple test for this that should work on 5.14+ and prior kernels.

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -42,6 +42,7 @@ import (
 const (
 	subtreeControl     = "cgroup.subtree_control"
 	controllersFile    = "cgroup.controllers"
+	killFile           = "cgroup.kill"
 	defaultCgroup2Path = "/sys/fs/cgroup"
 	defaultSlice       = "system.slice"
 )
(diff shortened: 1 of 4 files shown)


cgroup2/manager.go
```diff
@@ -366,6 +367,86 @@ func (c *Manager) AddThread(tid uint64) error {
...
+func (c *Manager) Kill() error {
+	v := Value{
+		filename: killFile,
+		value:    "1",
+	}
+	err := writeValues(c.path, []Value{v})
+	if err == nil {
+		return nil
+	}
+	logrus.Warnf("falling back to slower kill implementation: %s", err)
```
dcantah, line 384: Not sure this is worth a Warn level

AkihiroSuda (approved):

estesp (approved):
LGTM

PR #265, opened 2023-01-04
fix: return total system memory if the memory limit is not specified

ningziwen:
Signed-off-by: Ziwen Ning <ningziwe@amazon.com>

https://github.com/containerd/nerdctl/issues/1589

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,9 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/go-ole/go-ole v1.2.6 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/shirou/gopsutil v2.21.11+incompatible // indirect
+	github.com/yusufpapurcu/wmi v1.2.2 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/cgroup2/utils.go
+++ b/cgroup2/utils.go
@@ -32,6 +32,8 @@ import (
 
 	"github.com/godbus/dbus/v5"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/shirou/gopsutil/mem"
+
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 )
@@ -250,7 +252,14 @@ func getStatFileContentUint64(filePath string) uint64 {
 	}
 	trimmed := strings.TrimSpace(string(contents))
 	if trimmed == "max" {
-		return math.MaxUint64
+		totalMem, err := readTotalMem()
+		if err != nil {
+			// Reading total system memory adds a new dependency for a non-critical improvement.
+			// Fall back to the behaviour if there is anything wrong with system memory.
+			logrus.Warnf("falling back to returning hardcoded memory limit: %s", err)
+			return math.MaxUint64
+		}
+		return totalMem
 	}
 
 	res, err := parseUint(trimmed, 10, 64)
@@ -260,6 +269,14 @@ func getStatFileContentUint64(filePath string) uint64 {
 	}
 
 	return res
+}
+
+func readTotalMem() (uint64, error) {
+	vmStat, err := mem.VirtualMemory()
+	if err != nil {
+		return 0, err
+	}
+	return vmStat.Total, nil
 }
 
 func readIoStats(path string) []*stats.IOEntry {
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -17,6 +17,9 @@ require (
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/go-ole/go-ole v1.2.6 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/shirou/gopsutil v2.21.11+incompatible // indirect
```
estesp, line 22: `gopsutil` is a direct dependency so should be in the first `require` clause; `go mod tidy` should have put this in the correct place in `go.mod`; either way, that is why the CI project check failed.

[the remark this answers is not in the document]
ningziwen: Thanks! Fixed. I'm still figuring out how to test it and how to fix v1 so it is in draft.

go.mod
```diff
@@ -8,15 +8,22 @@ require (
 	github.com/docker/go-units v0.4.0
 	github.com/godbus/dbus/v5 v5.0.4
 	github.com/opencontainers/runtime-spec v1.0.2
+	github.com/shirou/gopsutil/v3 v3.22.12
```
kzys, line 11: Is it hard to implement by ourselves? cgroups is Linux-only whereas gopsutil supports other OSes and bring a bunch of dependencies.

kzys:
1. I prefer to not have this type of [Do What I Mean](https://en.wikipedia.org/wiki/DWIM) behavior in low-level packages. There may be customers we want to check whether the limit is configured or not. The current behavior (returning MaxUint64) may not be the best, but the new behavior is even harder for them.
2. There are other limits we can set on cgroups (see https://github.com/containerd/cgroups/search?q=limit). Would we do the same for others?

ningziwen:
> 1. I prefer to not have this type of [Do What I Mean](https://en.wikipedia.org/wiki/DWIM) behavior in low-level packages. There may be customers who want to check whether the limit is configured or not. The current behavior (returning MaxUint64) may not be the best, but the new behavior is even harder for them.
> 2. There are other limits we can set on cgroups (see https://github.com/containerd/cgroups/search?q=limit). Would we do the same for others?

@kzys 

If we change this in containerd, there is not a perfect way to check if the memory is configured. Returning Math.max memory can not only mean the memory is not configured, it can also mean the memory is really set to Math.max (that may be not possible today but may cause issue one day later). Do you think it is ok to use Math.max to check in containerd?

Similarily, if other consumers are using Math.max to judge if the memory is configured, they are currently having same defects. 

So there are a few options below:

* Option 1 change in containerd and keep cgroup same.
Pros: 
  * lowest risk to impact other consumers of cgroup

Cons: 
  * not perfectly logical to use Math.max to check if memory is configured in containerd, and other cgroup consuemrs. May cause issues in the future.
 
* Option 2 Change default memory limit to host memory limit.
Pros:
  * Consumers don't have to handle the default memory limit overwriting.
Cons:
  * May break existing consumer's workflow.

* Option 3 Change default memory limit to NA in cgroup layer. Modify to other
Pros:
  * Consumer's conditional check will be logical.
Cons:
  * May break existing consumer's workflow.

* Option 4 Keep cgroup same. Do some refactoring to add additional response field to tell the consumers that the memory limit is not set.
Pros:
  * Won't break any consumers.
  * Consumer's conditional check will be logical.
Cons:
  * Cgroup will need to return an additional field in the Stat() response.

I'm slightly leaning to Option 4. We can just add a field around here to show if the memory limit is configured. https://github.com/containerd/cgroups/blob/157c4da5ef3aac5fd7229b5ff3e13a8c0ee857d4/cgroup2/stats/metrics.pb.go#L297
(Similar for cgroup v1)

For the second question, I think we can apply the same option for other limits to keep them consistent.

Let me know if you have better ideas.

dcantah:
I'll give some thought to what might make sense to return here as well, but for the introduction of the new dependency it really should not be needed. It just reads `/proc/meminfo`, splits by line and then we just need the MemTotal section:

```
dcantah@dcantah ~ $ cat /proc/meminfo
MemTotal:       16338456 kB
MemFree:         5120496 kB
MemAvailable:   11223128 kB
Buffers:          889596 kB
... omitted for brevity
```

kzys:
I like Option 1 or Option 4.

@AkihiroSuda What do you think?

ningziwen:
@AkihiroSuda WDYT?

AkihiroSuda:
> Returning Math.max memory can not only mean the memory is not configured, it can also mean the memory is really set to Math.max (that may be not possible today but may cause issue one day later). Do you think it is ok to use Math.max to check in containerd?

I think this is fine.
Nobody will actually set the cgroup memory limit to 16EB until the ecosystem migrates to an 128-bit arch with >= 64-bit physical address bus, which isn’t realistic in this century.

AkihiroSuda:
But I agree Option 4 looks better

AkihiroSuda:
> We can just add a field around here to show if the memory limit is configured.

Can we just use zero or nil value to indicate that without adding a new field?

PR #268, opened 2023-01-30
gof(ump)t code and update some dependencies

thaJeztah:
Had this in draft locally, and looks like I never pushed 😅 

- gof(ump)t code
- go.mod: tidy with go1.18
- go.mod: golang.org/x/sys v0.2.0
  updating to use tagged releases
- go.mod: github.com/sirupsen/logrus v1.9.0
  update to current version, which is what's used by most projects already
- go.mod: github.com/cilium/ebpf v0.9.3
- cmd: go.mod: github.com/urfave/cli v1.22.12
   removes / updates some obsolete dependencies.


thaJeztah:
I think `github.com/pmezard/go-difflib v1.0.0` will be gone once we update to a newer ebpf, but it looks like there may be some (potentially) breaking changes in ebpf v0.10.0, so I didn't push forward to that (yet).

AkihiroSuda (approved):

dcantah (approved):

PR #269, opened 2023-02-06
cg2: syscall -> x/sys/unix

dcantah:
Change to x/sys/unix definitions. I left cgroups 1 as right now it *should* be able to build on non-linux platforms, but there's a singular `syscall.EINVAL` check which swapping to unix would ruin. cg2 should already not be able to build on non-linux as we were using `InotifyInit` and `Statfs_t` in tests.

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -27,7 +27,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


austinvazquez (approved):
Oh neat, TIL. LGTM

AkihiroSuda (approved):

PR #278, opened 2023-04-01
Reduce allocs in ReadUint64 by pre-allocating byte buffer

manugupt1:
PR captures the reasoning behind this:
https://github.com/containerd/cgroups/pull/275/files#r1152562543

    Reduce allocs in ReadUint64 by pre-allocating byte buffer
    
    Before
    
    Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkReaduint64$ github.com/containerd/cgroups/v3/cgroup1 -count=1 -timeout 30m
    
    goos: linux
    goarch: amd64
    pkg: github.com/containerd/cgroups/v3/cgroup1
    BenchmarkReaduint64-8             105118              9762 ns/op             848 B/op          5 allocs/op
    PASS
    ok      github.com/containerd/cgroups/v3/cgroup1        1.151s
    
    After
    
    Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkReaduint64$ github.com/containerd/cgroups/v3/cgroup1 -count=1 -timeout 30m
    
    goos: linux
    goarch: amd64
    pkg: github.com/containerd/cgroups/v3/cgroup1
    BenchmarkReaduint64-8             141484              8322 ns/op             128 B/op          3 allocs/op
    PASS
    ok      github.com/containerd/cgroups/v3/cgroup1        1.274s

### DIFF
--- a/cgroup1/utils.go
+++ b/cgroup1/utils.go
@@ -131,11 +131,19 @@ func hugePageSizes() ([]string, error) {
 }
 
 func readUint(path string) (uint64, error) {
-	v, err := os.ReadFile(path)
+	f, err := os.Open(path)
 	if err != nil {
 		return 0, err
 	}
-	return parseUint(strings.TrimSpace(string(v)), 10, 64)
+	defer f.Close()
+
+	b := make([]byte, 128) // Chose 128 as some files have leading/trailing whitespaces and alignment
+	n, err := f.Read(b)
+	if err != nil {
+		return 0, err
+	}
+
+	return parseUint(strings.TrimSpace(string(b[:n])), 10, 64)
 }
 
 func parseUint(s string, base, bitSize int) (uint64, error) {
--- a/cgroup1/utils_test.go
+++ b/cgroup1/utils_test.go
@@ -1,0 +1,32 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cgroup1
+
+import (
+	"testing"
+)
+
+func BenchmarkReaduint64(b *testing.B) {
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		_, err := readUint("/proc/self/loginuid")
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+}


manugupt1:
Cc @fuweid @dcantah @AkihiroSuda

cgroup1/utils.go
```diff
@@ -131,11 +133,19 @@ func hugePageSizes() ([]string, error) {
...
-	return parseUint(strings.TrimSpace(string(v)), 10, 64)
+	defer f.Close()
+	b := new(bytes.Buffer)
+	b.Reset()
+	_, err = io.Copy(b, f)
+	if err != nil {
+		return 0, err
+	}
+
+	return parseUint(strings.TrimSpace(b.String()), 10, 64)
```
dcantah, line 148: Seems just using a raw byte slice cuts allocs down to 4 (although I get 7 allocs on my machine while testing your change instead of 6 as you posted). Doesn't need to be 80 bytes either, but > 64 might be necessary seeming as how this code took precautions to trim the string, so I'm assuming there's some cases where there may be trailing space in these files. 

```go
f, err := os.Open(path)
if err != nil {
	return 0, err
}
defer f.Close()

buf := make([]byte, 80)
n, err := f.Read(buf)
if err != nil {
	return 0, err
}

return parseUint(strings.TrimSpace(string(buf[:n])), 10, 64)
```

dcantah:
```shell
goos: linux
goarch: amd64
pkg: github.com/containerd/cgroups/v3/cgroup1
cpu: AMD Ryzen 9 5900X 12-Core Processor            
BenchmarkReaduint64-24            849498              1342 ns/op             136 B/op          4 allocs/op
PASS
ok      github.com/containerd/cgroups/v3/cgroup1        1.160s
```

manugupt1: yeah! It is significantly lower. I used buffer because of TrimSpace but adding extra bytes works as well.
I think our benchmarks are different as well (mine ends with -8 and yours with -24); maybe that is why allocs are different.

manugupt1: I think we can benchmark bytebuffer in places where we are doing Readfile in entire package and see if we can reduce allocations in follow up PRs.

dcantah: That's fine with me

cgroup1/utils.go
```diff
@@ -131,11 +131,18 @@ func hugePageSizes() ([]string, error) {
...
 	if err != nil {
 		return 0, err
 	}
-	return parseUint(strings.TrimSpace(string(v)), 10, 64)
+	defer f.Close()
+	b := make([]byte, 80)
+	n, err := f.Read(b)
+	if err != nil {
+		return 0, err
+	}
```
dcantah, line 143: nit: Can you add an empty line between the f.Close and make of the byte buf

dcantah (approved):
LGTM (just one suggestion)

cgroup1/utils.go
```diff
@@ -131,11 +131,18 @@ func hugePageSizes() ([]string, error) {
...
 
 func readUint(path string) (uint64, error) {
-	v, err := os.ReadFile(path)
+	f, err := os.Open(path)
 	if err != nil {
 		return 0, err
 	}
-	return parseUint(strings.TrimSpace(string(v)), 10, 64)
+	defer f.Close()
+	b := make([]byte, 80)
```
dcantah, line 139: Actually one more comment. Perhaps we should put a note above this for the reasoning behind 80 e.g. arbitrary value above 64 to account for trailing space (most I remember seeing was just one space but 🤷‍♂️), should fit on stack, extra 16 from 64 shouldn't hurt anyone etc.

manugupt1: PTAL, if the comment needs more modification. Can you add a suggestion/the entire wording; so that I can commit directly in.

cgroup1/utils_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"testing"
+)
+
+func BenchmarkReaduint64(b *testing.B) {
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		_, err := readUint("testdata/uint")
```
kolyshkin, line 27: You forgot to commit testdata/uint file.

Alternatively, you can have the test create the temp file and then do `b.Reset()`.

dcantah: Wait, I see the test file committed? But +1 to just making a temp file here in place of that

kolyshkin:
> Before
> 
> goos: linux goarch: amd64 pkg: github.com/containerd/cgroups/v3/cgroup1 cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz BenchmarkPids-8 59758 19905 ns/op 2064 B/op 15 allocs/op

Not sure why you use `BenchmarkPids` before and `BenchmarkReaduint64` after.

cgroup1/utils_test.go
kolyshkin: Ah, sorry, missed that.

kolyshkin:
Technically speaking, the benchmark should use a real cgroupfs file (or any other file from /proc). The reason is, normal files have size, and `os.ReadFile` uses that information to prepare the buffer of the size needed. OTOH many cgroupfs (and /proc) files report the size of `0`, which makes os.ReadFile to allocate 512 bytes buffer.

I would recommend reading something like `/proc/self/loginuid`.

kolyshkin:
I see number of allocations going down from 5 to 3, when using Go 1.20.2 (or 6 to 4 when using Go 1.19.6). This is far from savings reported in the PR description (which is obviously wrong as it says `BenchmarkPid` for "before" measurement).

Out of these 2 allocations, one comes from the `f.Stat` call from `os.OpenFile`. Indeed, we don't need `stat()` when working with cgroup files. The other, I guess, comes from inlining (if I move file reading to a separate function, the number of allocs goes up from 3 to 4).

So yes, the commit makes sense, the description though needs to be modified to
 - show the correct before and after measurements
 - tell that we avoid os.OpenFile because:
   - it calls os.Stat which we do not need
   - it pre-allocates 512 bytes while we're fine with 80

In addition, I would change 80 to 128 for alignment.

kolyshkin:
Finally, note that there are a few other implementations in this repo, that are very similar to `readUint`. They all are in cgroup2 package, I was able to find two:
 - `getStatFileContentUint64` (uses `os.ReadFile`)
 - `readSingleFile` (uses `io.ReadAll`)
 
 I think those needs to be either consolidated, or fixed in a similar manner. Otherwise, we'll have the same problem in a couple of years once everyone has switched to cgroup v2.

dcantah:
@kolyshkin Oop, you beat me to your last comment. I was drafting an issue for similar fixes we could make here, have the same two you found so far, may be more.

dcantah:
> Technically speaking, the benchmark should use a real cgroupfs file (or any other file from /proc). The reason is, normal files have size, and os.ReadFile uses that information to prepare the buffer of the size needed. OTOH many cgroupfs (and /proc) files report the size of 0, which makes os.ReadFile to allocate 512 bytes buffer.

Probably is better to test on some magic file so +1, but given the change here is for us to provide the buffer, the `ReadFile` behavior doesn't come into the picture right?

manugupt1:
Yes to all of the above. I will do follow ups in coming days

kolyshkin:
> Probably is better to test on some magic file so +1, but given the change here is for us to provide the buffer, the `ReadFile` behavior doesn't come into the picture right?

The `ReadFile` is used in the "before" benchmark.

dcantah:
> > Probably is better to test on some magic file so +1, but given the change here is for us to provide the buffer, the `ReadFile` behavior doesn't come into the picture right?
> 
> The `ReadFile` is used in the "before" benchmark.

Ahh, I see what you were getting at now 👍.

manugupt1:
Sorry; I got mixed up while working. @kolyshkin is right! It is 5 --> 3

kolyshkin (approved):
LGTM, except we need to also address these: https://github.com/containerd/cgroups/pull/278#issuecomment-1496720308

dcantah (approved):
LGTM. @kolyshkin, I was working on fixing the two cases we found for cg2 (which reminds me I should open my issue here..)

brancz:
Ah awesome, I was going to get back to this today, glad you already worked it out! Re-running the benchmarks I now see a total improvement of ~25%!

```
$ benchstat old.txt new.txt
name          old time/op    new time/op    delta
Readuint64-8    5.19µs ± 2%    3.87µs ± 1%  -25.48%  (p=0.000 n=10+10)

name          old alloc/op   new alloc/op   delta
Readuint64-8      840B ± 0%      120B ± 0%  -85.71%  (p=0.000 n=10+10)

name          old allocs/op  new allocs/op  delta
Readuint64-8      5.00 ± 0%      3.00 ± 0%  -40.00%  (p=0.000 n=10+10)
```

That means we should reap a total improvement of about 4%! Awesome!

I'm going to go ahead and close my PR.

manugupt1:
Thanks for all the help @brancz and finding this issue.

fuweid (approved):
LGTM

Issue #280, opened 2023-04-05
Reduce allocations for Stat in cg2


dcantah:
There's been some recent PRs focusing on reducing allocs for cgroups1 with some wonderful work [here](https://github.com/containerd/cgroups/pull/275) and [here](https://github.com/containerd/cgroups/pull/278).

There's [some spots for cgroups2](https://github.com/containerd/cgroups/pull/278#issuecomment-1496720308) that could be improved also. Pasting inline:

- getStatFileContentUint64 ([uses os.ReadFile](https://github.com/containerd/cgroups/blob/49edf542aa03131efa6545a292f1deec4be5290e/cgroup2/utils.go#L245))
- readSingleFile ([uses io.ReadAll](https://github.com/containerd/cgroups/blob/49edf542aa03131efa6545a292f1deec4be5290e/cgroup2/manager.go#L662))

dcantah:
Gonna try and get to this by EOW

dcantah:
A large bulk of the allocations in Stat's path were removed in https://github.com/containerd/cgroups/pull/281. Be happy to re-open if we find any other pain points

PR #281, opened 2023-04-07
Cgroup2: Reduce allocations for `manager.Stat`

dcantah:
Related to https://github.com/containerd/cgroups/issues/280. There's still some spots we can fix, as the majority of allocs are occurring in bufio.Scan in readKVStatsFile. cc @kolyshkin @manugupt1 seeming as how we were all there for the last one 

This change works towards bringing down allocations for manager.Stat(). There's quite a few things we can do:

1. Swap to opening a file and issuing a single read for uint64/"max" values. Previously we were doing os.ReadFile's which returns a byte slice, so it needs to allocate.
2. Sizing the map we store {controller}.stat values in. We store 40+ stats in this map and the default bucket size for Go seems to be smaller than this, so we'd incur a couple allocs at runtime when adding these.
3. Finally, and funnily enough the biggest win, was just not calling f.Readdir(-1) for hugeTlb stats, and swapping this for a trick the runc cgroup code does. The number of allocs shaved off here is tied to the number of files in the cgroup directory
as f.Readdir(-1) was being called everytime we went to collect HugeTlb stats (it returns a slice of interfaces, that all we were doing is using the Name() method of anyways). To optimize this we can take advantage of the fact that we know exactly what files we need to open as they're fixed, the only variable portion is the hugepage sizes available on the host. To get around this we can compute the hugepage sizes in a sync.Once, and all future manager.Stat calls can reap the benefits as we don't need to Readdir() the entire cgroup path to search for hugetlb.pagesize.foobar's.

All in all on Go 1.19.4 this drops allocations in Stat from 322 to 198

```shell
benchstat ~/benchstat/old.txt ~/benchstat/new.txt 
goos: linux
goarch: amd64
pkg: github.com/containerd/cgroups/v3/cgroup2
cpu: AMD Ryzen 9 5900X 12-Core Processor            
        │ /home/dcantah/benchstat/old.txt │   /home/dcantah/benchstat/new.txt   │
        │             sec/op              │   sec/op     vs base                │
Stat-24                      112.29µ ± 3%   65.10µ ± 6%  -42.02% (p=0.000 n=10)

        │ /home/dcantah/benchstat/old.txt │   /home/dcantah/benchstat/new.txt    │
        │              B/op               │     B/op      vs base                │
Stat-24                      41.40Ki ± 0%   22.26Ki ± 0%  -46.23% (p=0.000 n=10)

        │ /home/dcantah/benchstat/old.txt │  /home/dcantah/benchstat/new.txt   │
        │            allocs/op            │ allocs/op   vs base                │
Stat-24                        322.0 ± 0%   198.0 ± 0%  -38.51% (p=0.000 n=10)
```

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -523,7 +523,10 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 	if err != nil {
 		return nil, err
 	}
-	out := make(map[string]uint64)
+	// Sizing this avoids an allocation to increase the map at runtime;
+	// currently the default bucket size is 8 and we put 40+ elements
+	// in it so we'd always end up allocating.
+	out := make(map[string]uint64, 50)
 	for _, controller := range controllers {
 		switch controller {
 		case "cpu", "memory":
@@ -541,8 +544,8 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 			return nil, err
 		}
 	}
-	var metrics stats.Metrics
 
+	var metrics stats.Metrics
 	metrics.Pids = &stats.PidsStat{
 		Current: getStatFileContentUint64(filepath.Join(c.path, "pids.current")),
 		Limit:   getStatFileContentUint64(filepath.Join(c.path, "pids.max")),
--- a/cgroup2/manager_test.go
+++ b/cgroup2/manager_test.go
@@ -223,3 +223,20 @@ func TestCgroupType(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, cgType, Threaded)
 }
+
+func BenchmarkStat(b *testing.B) {
+	checkCgroupMode(b)
+	group := "/stat-test-cg"
+	groupPath := fmt.Sprintf("%s-%d", group, os.Getpid())
+	c, err := NewManager(defaultCgroup2Path, groupPath, &Resources{})
+	require.NoErrorf(b, err, "failed to init new cgroup manager")
+	b.Cleanup(func() {
+		_ = c.Delete()
+	})
+
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		_, err := c.Stat()
+		require.NoError(b, err)
+	}
+}
--- a/cgroup2/testutils_test.go
+++ b/cgroup2/testutils_test.go
@@ -27,14 +27,14 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func checkCgroupMode(t *testing.T) {
+func checkCgroupMode(tb testing.TB) {
 	var st unix.Statfs_t
 	err := unix.Statfs(defaultCgroup2Path, &st)
-	require.NoError(t, err, "cannot statfs cgroup root")
+	require.NoError(tb, err, "cannot statfs cgroup root")
 
 	isUnified := st.Type == unix.CGROUP2_SUPER_MAGIC
 	if !isUnified {
-		t.Skip("System running in hybrid or cgroupv1 mode")
+		tb.Skip("System running in hybrid or cgroupv1 mode")
 	}
 }
 
--- a/cgroup2/utils.go
+++ b/cgroup2/utils.go
@@ -18,6 +18,7 @@ package cgroup2
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -25,6 +26,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 	"unsafe"
 
@@ -387,56 +389,94 @@ func systemdUnitFromPath(path string) string {
 }
 
 func readHugeTlbStats(path string) []*stats.HugeTlbStat {
-	usage := []*stats.HugeTlbStat{}
-	keyUsage := make(map[string]*stats.HugeTlbStat)
-	f, err := os.Open(path)
-	if err != nil {
-		return usage
-	}
-	files, err := f.Readdir(-1)
-	f.Close()
-	if err != nil {
-		return usage
-	}
-
-	for _, file := range files {
-		if strings.Contains(file.Name(), "hugetlb") &&
-			(strings.HasSuffix(file.Name(), "max") || strings.HasSuffix(file.Name(), "current")) {
-			var hugeTlb *stats.HugeTlbStat
-			var ok bool
-			fileName := strings.Split(file.Name(), ".")
-			pageSize := fileName[1]
-			if hugeTlb, ok = keyUsage[pageSize]; !ok {
-				hugeTlb = &stats.HugeTlbStat{}
-			}
-			hugeTlb.Pagesize = pageSize
-			out, err := os.ReadFile(filepath.Join(path, file.Name()))
-			if err != nil {
-				continue
-			}
-			var value uint64
-			stringVal := strings.TrimSpace(string(out))
-			if stringVal == "max" {
-				value = math.MaxUint64
-			} else {
-				value, err = strconv.ParseUint(stringVal, 10, 64)
-			}
-			if err != nil {
-				continue
-			}
-			switch fileName[2] {
-			case "max":
-				hugeTlb.Max = value
-			case "current":
-				hugeTlb.Current = value
-			}
-			keyUsage[pageSize] = hugeTlb
+	hpSizes := hugePageSizes()
+	usage := make([]*stats.HugeTlbStat, len(hpSizes))
+	for idx, pagesize := range hpSizes {
+		usage[idx] = &stats.HugeTlbStat{
+			Max:      getStatFileContentUint64(filepath.Join(path, "hugetlb."+pagesize+".max")),
+			Current:  getStatFileContentUint64(filepath.Join(path, "hugetlb."+pagesize+".current")),
+			Pagesize: pagesize,
 		}
 	}
-	for _, entry := range keyUsage {
-		usage = append(usage, entry)
-	}
 	return usage
+}
+
+var (
+	hPageSizes  []string
+	initHPSOnce sync.Once
+)
+
+// The following idea and implementation is taken pretty much line for line from
+// runc. Because the hugetlb files are well known, and the only variable thrown in
+// the mix is what huge page sizes you have on your host, this lends itself well
+// to doing the work to find the files present once, and then re-using this. This
+// saves a os.Readdirnames(0) call to search for hugeltb files on every `manager.Stat`
+// call.
+// https://github.com/opencontainers/runc/blob/3a2c0c2565644d8a7e0f1dd594a060b21fa96cf1/libcontainer/cgroups/utils.go#L301
+func hugePageSizes() []string {
+	initHPSOnce.Do(func() {
+		dir, err := os.OpenFile("/sys/kernel/mm/hugepages", unix.O_DIRECTORY|unix.O_RDONLY, 0)
+		if err != nil {
+			return
+		}
+		files, err := dir.Readdirnames(0)
+		dir.Close()
+		if err != nil {
+			return
+		}
+
+		hPageSizes, err = getHugePageSizeFromFilenames(files)
+		if err != nil {
+			logrus.Warnf("hugePageSizes: %s", err)
+		}
+	})
+
+	return hPageSizes
+}
+
+func getHugePageSizeFromFilenames(fileNames []string) ([]string, error) {
+	pageSizes := make([]string, 0, len(fileNames))
+	var warn error
+
+	for _, file := range fileNames {
+		// example: hugepages-1048576kB
+		val := strings.TrimPrefix(file, "hugepages-")
+		if len(val) == len(file) {
+			// Unexpected file name: no prefix found, ignore it.
+			continue
+		}
+		// In all known versions of Linux up to 6.3 the suffix is always
+		// "kB". If we find something else, produce an error but keep going.
+		eLen := len(val) - 2
+		val = strings.TrimSuffix(val, "kB")
+		if len(val) != eLen {
+			// Highly unlikely.
+			if warn == nil {
+				warn = errors.New(file + `: invalid suffix (expected "kB")`)
+			}
+			continue
+		}
+		size, err := strconv.Atoi(val)
+		if err != nil {
+			// Highly unlikely.
+			if warn == nil {
+				warn = fmt.Errorf("%s: %w", file, err)
+			}
+			continue
+		}
+		// Model after https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/hugetlb_cgroup.c?id=eff48ddeab782e35e58ccc8853f7386bbae9dec4#n574
+		// but in our case the size is in KB already.
+		if size >= (1 << 20) {
+			val = strconv.Itoa(size>>20) + "GB"
+		} else if size >= (1 << 10) {
+			val = strconv.Itoa(size>>10) + "MB"
+		} else {
+			val += "KB"
+		}
+		pageSizes = append(pageSizes, val)
+	}
+
+	return pageSizes, warn
 }
 
 func getSubreaper() (int, error) {
--- a/cgroup2/utils.go
+++ b/cgroup2/utils.go
@@ -236,18 +236,28 @@ func ToResources(spec *specs.LinuxResources) *Resources {
 
 // Gets uint64 parsed content of single value cgroup stat file
 func getStatFileContentUint64(filePath string) uint64 {
-	contents, err := os.ReadFile(filePath)
+	f, err := os.Open(filePath)
 	if err != nil {
 		return 0
 	}
-	trimmed := strings.TrimSpace(string(contents))
+	defer f.Close()
+
+	// We expect an unsigned 64 bit integer, or a "max" string
+	// in some cases.
+	buf := make([]byte, 32)
+	n, err := f.Read(buf)
+	if err != nil {
+		return 0
+	}
+
+	trimmed := strings.TrimSpace(string(buf[:n]))
 	if trimmed == "max" {
 		return math.MaxUint64
 	}
 
 	res, err := parseUint(trimmed, 10, 64)
 	if err != nil {
-		logrus.Errorf("unable to parse %q as a uint from Cgroup file %q", string(contents), filePath)
+		logrus.Errorf("unable to parse %q as a uint from Cgroup file %q", trimmed, filePath)
 		return res
 	}
 
--- a/cgroup2/utils_test.go
+++ b/cgroup2/utils_test.go
@@ -58,3 +58,11 @@ func TestToResources(t *testing.T) {
 	v2resources2 := ToResources(&res2)
 	assert.Equal(t, CPUMax("max 10000"), v2resources2.CPU.Max)
 }
+
+func BenchmarkGetStatFileContentUint64(b *testing.B) {
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		_ = getStatFileContentUint64("/proc/self/loginuid")
+	}
+}


dcantah:
Working towards https://github.com/containerd/cgroups/issues/280. There's still some spots we can fix, as the majority of allocs are occurring in bufio.Scan in readKVStatsFile

dcantah:
cc @kolyshkin @manugupt1 seeming as how we were all there for the last one 🤣

cgroup2/manager.go
```diff
@@ -763,37 +755,42 @@ func (c *Manager) EventChan() (<-chan Event, <-chan error) {
...
 
-func parseMemoryEvents(out map[string]interface{}) (Event, error) {
+func parseMemoryEvents(out map[string]string) (Event, error) {
 	e := Event{}
 	if v, ok := out["high"]; ok {
-		e.High, ok = v.(uint64)
-		if !ok {
+		val, err := parseUint(v, 10, 64)
+		if err != nil {
 			return Event{}, fmt.Errorf("cannot convert high to uint64: %+v", v)
```
kolyshkin, line 763: Not related to this PR, but it would be easier to return `nil`not `Event{}` every time an error is returned

cgroup2/utils_test.go
```diff
@@ -64,3 +64,22 @@ func TestToResources(t *testing.T) {
...
+		_ = getStatFileContentUint64("/proc/self/loginuid")
+	}
+}
+
+func BenchmarkReadSingleFile(b *testing.B) {
+	b.ReportAllocs()
+
+	out := make(map[string]string)
+	for i := 0; i < b.N; i++ {
+		if err := readSingleFile("/proc/self/loginuid", "", out); err != nil {
```
kolyshkin, line 81: readSingleFile expects a file with key value pairs, so `/proc/self/loginuid` is not a good choice.

cgroup2/utils.go
```diff
@@ -242,21 +237,30 @@ func ToResources(spec *specs.LinuxResources) *Resources {
...
 		return math.MaxUint64
 	}
 
 	res, err := parseUint(trimmed, 10, 64)
 	if err != nil {
-		logrus.Errorf("unable to parse %q as a uint from Cgroup file %q", string(contents), filePath)
+		logrus.Errorf("unable to parse %q as a uint from Cgroup file %q", trimmed, filePath)
 		return res
 	}
-
```
kolyshkin, line 259: this should not be here

kolyshkin:
I suggest making `readHugeTlbStats` changes related to readdirnames as a separate commit.

dcantah:
> I suggest making `readHugeTlbStats` changes related to readdirnames as a separate commit.

That's fair, they were all in the pursuit of improving Stat so I felt either way. I originally had the readSingleFile and uint improvements in a separate commit and squished after finding the other areas. I'll separate them though

cgroup2/utils_test.go
dcantah: It doesn't afaict, it just places the resulting value in the passed in map. The original logic was to try and parse the output as a uint, and if it failed just treat the output as a string. It's only called for pids.current and pids.max which are both either "max" or some unsigned integer

cgroup2/manager.go
dcantah: Agree, I'm going to leave though to avoid too much refactor off the main goal. We can always do a followup if need be

cgroup2/utils.go
dcantah: Ack, fixing in a little!

dcantah:
@kolyshkin Made a separate commit here for the hugeTlb changes, thanks for the quick review!

cgroup2/manager.go
```diff
@@ -529,7 +528,10 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 	if err != nil {
 		return nil, err
 	}
-	out := make(map[string]interface{})
+	// Sizing this avoids an allocation to increase the map at runtime;
+	// currently the default bucket size is 8 and we put 40+ elements
+	// in it so we'd always end up allocating.
+	out := make(map[string]string, 50)
```
kolyshkin, line 534: perhaps have the size as 0 and capacity as 64?

cgroup2/utils_test.go
kolyshkin: Sorry, I mixed that up with readKVStatsFile. Yes, this is the right approach then.

cgroup2/manager.go
dcantah: You can't set the capacity of a map or chan, only slices

kolyshkin:
By looking at the code, I'd rather have it refactored first, then optimize. At least some parts that are crying to be removed, such as readSingleFile -- PTAL https://github.com/containerd/cgroups/pull/283

dcantah:
@kolyshkin ack, there is quite a few places that all do the same thing. I'm fine getting rid of them and then rebasing here

cgroup2/manager.go
kolyshkin: Right, my bad, ENEEDCOFFEE

cgroup2/manager.go
```diff
@@ -625,56 +627,53 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
...
 	f, err := os.Open(filepath.Join(path, file))
 	if err != nil {
 		return err
 	}
 	defer f.Close()
-	data, err := io.ReadAll(f)
+
+	// We expect an unsigned 64 bit integer, or a "max" string
+	// in some cases.
+	buf := make([]byte, 32)
```
manugupt1, line 666: Is this way too low or
https://github.com/containerd/cgroups/pull/278/files#diff-f6fc333a9f621c02b1a9723d8cae6ad3ddd6cdc2501c75cc2a5765bbf01e631cR140 this too high?

cgroup2/utils_test.go
```diff
@@ -64,3 +64,22 @@ func TestToResources(t *testing.T) {
 	v2resources2 := ToResources(&res2)
 	assert.Equal(t, CPUMax("max 10000"), v2resources2.CPU.Max)
 }
+
+func BenchmarkReaduint64(b *testing.B) {
```
manugupt1, line 68: nit: BenahcMarkGetStatFileContentUint64

cgroup2/manager.go
dcantah: Yea, I think it's too high. It was originally a brainfart on my part of attributing 64 bit with needing 64 bytes for the character representation in the file, and I found out what the trimming of space was likely for for all of these parsing functions. All of the cgroup files seem to end with a single line feed, and TrimSpace counts this as white space, so it's basically just removing \n for any of the files.  We should only need a buffer of 20 to handle the max uint64, so 32's just for power of 2. I'll open a change to lower the other buffer

dcantah:
Drafting this till https://github.com/containerd/cgroups/pull/283 gets in

cgroup2/utils.go
```diff
@@ -379,41 +389,55 @@ func systemdUnitFromPath(path string) string {
...
 	usage := []*stats.HugeTlbStat{}
 	keyUsage := make(map[string]*stats.HugeTlbStat)
+
 	f, err := os.Open(path)
 	if err != nil {
 		return usage
 	}
-	files, err := f.Readdir(-1)
+
+	files, err := f.Readdirnames(-1)
```
manugupt1, line 398: You can change to improve further marginally by using f.ReadDir(-1).
See: https://go.dev/play/p/aMmQKCKBZpN

I got 
```
-timeout 30m

goos: linux
goarch: amd64
pkg: github.com/manugupt1/file
BenchmarkReaddir-8        	     871	   1300868 ns/op	  130113 B/op	    1439 allocs/op
BenchmarkReadDir-8        	    3076	    370488 ns/op	   48972 B/op	     957 allocs/op
BenchmarkReadDirNames-8   	    3139	    366333 ns/op	   48973 B/op	     957 allocs/op
PASS
	github.com/manugupt1/file	coverage: 50.0% of statements
ok  	github.com/manugupt1/file	3.655s
```

cgroup2/utils.go
```diff
@@ -379,41 +389,55 @@ func systemdUnitFromPath(path string) string {
...
 				hugeTlb = &stats.HugeTlbStat{}
 			}
 			hugeTlb.Pagesize = pageSize
-			out, err := os.ReadFile(filepath.Join(path, file.Name()))
+
+			f, err := os.Open(filepath.Join(path, file))
 			if err != nil {
 				continue
 			}
+			defer f.Close()
```
manugupt1, line 423: nit: Can/Should we instead try to close the file pointer in the function after reading rather than defering the call?

cgroup2/utils.go
dcantah: Hmm, this is what I swapped to originally but found the latter shaved down a few more allocations. Let me try again

manugupt1: Hmm! I am not sure why; maybe a typo : ReadDir vs Readdir, because ReadDir is further down the stack when you look through go's source code.

cgroup2/utils.go
```diff
@@ -379,41 +389,55 @@ func systemdUnitFromPath(path string) string {
...
+
+			n, err := f.Read(buf)
+			if err != nil {
+				continue
+			}
+
 			var value uint64
-			stringVal := strings.TrimSpace(string(out))
+			stringVal := strings.TrimSpace(string(buf[:n]))
```
manugupt1, line 431: Another realization: if we use the bytes package; its again faster; does not change allocs though.
See: https://go.dev/play/p/7jPtYNfN7xq

I get 
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^(BenchmarkByteCompare|BenchmarkStringCompare)$ github.com/manugupt1/file -count=1 -timeout 30m

goos: linux
goarch: amd64
pkg: github.com/manugupt1/file
BenchmarkByteCompare-8     	64981508	        17.42 ns/op	       0 B/op	       0 allocs/op
BenchmarkStringCompare-8   	41275358	        27.36 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/manugupt1/file	3.223s
```

manugupt1: I will do these for cgroup1 if you are fine with it.

cgroup2/utils.go
dcantah: No it was definitely ReadDir that I tried as I noticed the stdlib comment on Readdir originally which sparked this investigation https://pkg.go.dev/os#File.Readdir:

"Most clients are better served by the more efficient ReadDir method." What Go version are you on? Should upgrade to 1.20 on my end on my linux machine 😳

manugupt1: Ah I am on go1.20 that's it

cgroup2/utils.go
dcantah: Go for it! That's a good point, we can defer the converting to strings after we trim

cgroup2/utils.go
dcantah: You mean manual f.Close calls if Read fails/if read succeeds?

cgroup2/utils.go
kolyshkin: I'm thinking maybe we should use an alternative approach here. The thing is, all the file names are well known ahead of time. The only thing that needs to be known is the list of hugetlb sizes (and for that, there's a function in cgroup1 package available (which needs to be wrapped in `once.Do`, but that's a different topic)).

When we know all page sizes, the code becomes something like this (an example from runc/libcontainer):

https://github.com/opencontainers/runc/blob/3a2c0c2565644d8a7e0f1dd594a060b21fa96cf1/libcontainer/cgroups/fs/hugetlb.go#L31-L59

dcantah: That does seem more ideal. I'll play with this tonight, thanks!

kolyshkin: As for the three versions of reading the file names,
- `os.Readdir` is the worst, because it calls `stat(2)` on every file;
- `os.ReadDir` is better as it doesn't do `stat` but instead wraps each entry into FileInfo interface;
- `os.Readdirnames` is the fastest as it just returns the list of names.

Since the name is everything we care about here, `os.Readdirnames` is the best.

Yet I would rewrite this using the approach without reading the directly (see above).

cgroup2/utils.go
```diff
@@ -389,41 +389,55 @@ func systemdUnitFromPath(path string) string {
...
 	}
 
+	buf := make([]byte, 32)
 	for _, file := range files {
-		if strings.Contains(file.Name(), "hugetlb") &&
-			(strings.HasSuffix(file.Name(), "max") || strings.HasSuffix(file.Name(), "current")) {
-			var hugeTlb *stats.HugeTlbStat
-			var ok bool
-			fileName := strings.Split(file.Name(), ".")
+		if strings.Contains(file, "hugetlb") &&
```
kolyshkin, line 406: While at it, makes sense to switch to `strings.HasPrefix`, and add a `.` after `hugetlb`.

cgroup2/utils.go
```diff
@@ -389,41 +389,55 @@ func systemdUnitFromPath(path string) string {
...
+			(strings.HasSuffix(file, "max") || strings.HasSuffix(file, "current")) {
```
kolyshkin, line 407: same here -- these better be `.max` and `.current`.

cgroup2/utils.go
dcantah: ack. I usually would leave things like this for another cleanup round but I'll just make another commit on top here. We're already screwing around in this area so it's worthwhile

cgroup2/utils.go
```diff
@@ -389,41 +389,55 @@ func systemdUnitFromPath(path string) string {
...
+			fileName := strings.Split(file, ".")
 			pageSize := fileName[1]
 			if hugeTlb, ok = keyUsage[pageSize]; !ok {
 				hugeTlb = &stats.HugeTlbStat{}
 			}
 			hugeTlb.Pagesize = pageSize
-			out, err := os.ReadFile(filepath.Join(path, file.Name()))
+
+			f, err := os.Open(filepath.Join(path, file))
 			if err != nil {
```
kolyshkin, line 420: Hmm, why not use `getStatFileContentUint64` from here?

dcantah: No reason other than forgetting 😆

cgroup2/utils.go
kolyshkin: @manugupt1 is right, because otherwise we may have quite a few files opened before we exit from this function.

Yet better -- switch to `getStatFileContentUint64`

dcantah: ^ That seems best

cgroup2/utils.go
manugupt1: I read wrong again :(

dcantah:
Okay, updated to include the runc trick @kolyshkin, great suggestion. We're down to 199 now.

cgroup2/utils.go
```diff
@@ -378,55 +390,92 @@ func systemdUnitFromPath(path string) string {
...
-	files, err := f.Readdir(-1)
-	f.Close()
-	if err != nil {
-		return usage
+	for _, pagesize := range hugePageSizes() {
+		usage = append(usage, &stats.HugeTlbStat{
+			Max:      getStatFileContentUint64(filepath.Join(path, "hugetlb."+pagesize+".max")),
+			Current:  getStatFileContentUint64(filepath.Join(path, "hugetlb."+pagesize+".current")),
+			Pagesize: pagesize,
+		})
```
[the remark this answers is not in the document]
manugupt1: Hmm! Can we do soemthing like
```
hugePageSizeSlice := hugePageSizes()
var usage = make(*stats.HugeTlbStat{}, len(hugePageSizeSlice)
and then iterate like an array
for idx, pageSize  := range hugePageSizeSlice {
  usage[idx] = .... 
}
```

cgroup2/manager_test.go
```diff
@@ -223,3 +223,22 @@ func TestCgroupType(t *testing.T) {
...
+
+func BenchmarkStat(b *testing.B) {
+	checkCgroupMode(b)
+	group := "/stat-test-cg"
+	groupPath := fmt.Sprintf("%s-%d", group, os.Getpid())
+	c, err := NewManager(defaultCgroup2Path, groupPath, &Resources{})
+	if err != nil {
+		b.Fatal("failed to init new cgroup manager: ", err)
+	}
+	defer os.Remove(c.path)
```
manugupt1, line 235: nit: b.Cleanup

cgroup2/testutils_test.go
```diff
@@ -27,14 +27,14 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func checkCgroupMode(t *testing.T) {
+func checkCgroupMode(tb testing.TB) {
 	var st unix.Statfs_t
-	err := unix.Statfs(defaultCgroup2Path, &st)
-	require.NoError(t, err, "cannot statfs cgroup root")
-
+	if err := unix.Statfs(defaultCgroup2Path, &st); err != nil {
```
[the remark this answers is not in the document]
dcantah: Oh that's weird, my editor was yelling at me originally when I tried putting `require.NoError(tb, err)` here, but trying on my personal pc works fine..? require takes in `require.TestingT`'s which is just an interface wrapper defined as:

```go
type TestingT interface {
	Errorf(format [string](https://pkg.go.dev/builtin#string), args ...interface{})
	FailNow()
}
```

and testing.TB has both defined so it should've always worked 🤷‍♂️. Changing

cgroup2/utils.go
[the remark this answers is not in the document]
dcantah: This is the same outcome though right? We still have a slice with empty entries, where the number of them is tied to how many huge page sizes we found in `/sys/kernel/mm/hugepages`. Before this we'd return an empty slice which I think is ideal

cgroup2/utils.go
```diff
@@ -378,55 +390,92 @@ func systemdUnitFromPath(path string) string {
...
+	var warn error
+
+	for _, file := range fileNames {
+		// example: hugepages-1048576kB
+		val := strings.TrimPrefix(file, "hugepages-")
+		if len(val) == len(file) {
+			// Unexpected file name: no prefix found, ignore it.
+			continue
+		}
+		// The suffix is always "kB" (as of Linux 5.13). If we find
```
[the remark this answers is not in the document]
dcantah: The comment is meant to mean as of 5.13 there's nothing but kB, not that 5.13 changed this to be all kB

cgroup2/utils.go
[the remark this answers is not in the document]
manugupt1: Oh okay! yes. what I suggested matches your current code; but I was looking from an allocation perspective. If you pre-allocate; the slice won't have to expand while looping and appending.

I do not have an opinion on functionality as I do not know how users use it. I would try and keep the previous behavior though.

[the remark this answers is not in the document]
kolyshkin: I agree with @manugupt1 that pre-allocating the `usage` slice is better than growing it dynamically. Even if it won't shave off any allocations, this is more optimized overall, and since we know the size, it makes total sense to do so.

cgroup2/utils.go
```diff
@@ -388,55 +390,92 @@ func systemdUnitFromPath(path string) string {
...
+		}
+		files, err := dir.Readdirnames(0)
+		dir.Close()
+		if err != nil {
+			return
+		}
+
+		hPageSizes, err = getHugePageSizeFromFilenames(files)
+		if err != nil {
+			logrus.Warnf("HugePageSizes: %s", err)
```
kolyshkin, line 429: nit: `s/Huge/huge/`

cgroup2/utils.go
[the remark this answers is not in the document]
kolyshkin: We can say something like "in all known Linux versions up to 6.3" (it's still always `kB` in there, I just checked).

kolyshkin (commented):
LGTM overall, except for preallocating the usage slice would be better.

@dcantah also, if you like, you can add `Co-authored-by: Kir Kolyshkin <kolyshkin@gmail.com>` to the second commit, as I wrote that venerable /sys/kernel/mm/hugepages/* parsing code.

kolyshkin:
> All in all on Go 1.19.4 this drops allocations in Stat from 322 to 199, or 47%.

Oh, and if you can use `benchstat` for such calculations that'd be terrific.

cgroup2/utils.go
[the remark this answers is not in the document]
dcantah: Oh I agree about pre-allocating, more the point I was trying to get across still stands: this behavior is different than whats here today. We'll have entries in hugetlb stats regardless of if someone even has the controller enabled, as long as there's entries in /sys/kernel/mm/hugepages

dcantah:
@kolyshkin Sure thing! I didn't even realize you wrote that, so you've already gone through this pain 🤣

cgroup2/utils.go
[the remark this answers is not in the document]
manugupt1: Right! I am not the best judge of this tbh!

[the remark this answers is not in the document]
dcantah: Lets see what others think, I don't think this is too crazy a departure

dcantah:
> Oh, and if you can use `benchstat` for such calculations that'd be terrific.

No wonder you recommended, I fat fingered 47 from 38.. No more programming past 12 🐸

fuweid (approved):
LGTM

estesp (approved):

PR #282, opened 2023-04-07
Cgroup2: Remove lingering cgroups after tests

dcantah:
There was a couple cgroup tests that wouldn't cleanup after themselves. This alters the names of these so they're more obvious and calls kill+removes the path if necessary.

### DIFF
--- a/cgroup2/manager_test.go
+++ b/cgroup2/manager_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"syscall"
 	"testing"
 	"time"
 
@@ -140,6 +141,9 @@ func TestKill(t *testing.T) {
 	checkCgroupMode(t)
 	manager, err := NewManager(defaultCgroup2Path, "/test1", ToResources(&specs.LinuxResources{}))
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		_ = manager.Delete()
+	})
 
 	var (
 		procs    []*exec.Cmd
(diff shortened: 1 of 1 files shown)


cgroup2/manager_test.go
```diff
@@ -194,21 +199,35 @@ func TestKill(t *testing.T) {
...
+	if err := cmd.Start(); err != nil {
+		t.Fatal(err)
+	}
+	proc := cmd.Process.Pid
 	if err := manager.AddProc(uint64(proc)); err != nil {
 		t.Error(err)
 		return
 	}
-	destination, err := NewManager(defaultCgroup2Path, "/test2", ToResources(&specs.LinuxResources{}))
+	destination, err := NewManager(defaultCgroup2Path, "/test-moveto-2", ToResources(&specs.LinuxResources{}))
```
manugupt1, line 222: nit: can we rename the paths as destination

cgroup2/manager_test.go
```diff
@@ -194,21 +199,35 @@ func TestKill(t *testing.T) {
...
 	checkCgroupMode(t)
-	manager, err := NewManager(defaultCgroup2Path, "/test1", ToResources(&specs.LinuxResources{}))
+
+	manager, err := NewManager(defaultCgroup2Path, "/test-moveto-1", ToResources(&specs.LinuxResources{}))
 	if err != nil {
 		t.Error(err)
 		return
 	}
-	proc := os.Getpid()
+	defer func() {
```
manugupt1, line 208: Can we use t.Cleanup
https://pkg.go.dev/testing#T.Cleanup

cgroup2/manager_test.go
```diff
@@ -230,8 +249,9 @@ func TestMoveTo(t *testing.T) {
 
 func TestCgroupType(t *testing.T) {
 	checkCgroupMode(t)
-	manager, err := NewManager(defaultCgroup2Path, "/test1", ToResources(&specs.LinuxResources{}))
+	manager, err := NewManager(defaultCgroup2Path, "/test-type", ToResources(&specs.LinuxResources{}))
 	require.NoError(t, err)
+	defer os.RemoveAll(manager.path)
```
manugupt1, line 254: same: can we use t.Cleanup

dcantah: ack, always forget this exists

cgroup2/manager_test.go
dcantah: You mean rename the destination variable NewManager returns or add destination somewhere in /test-moveto?

cgroup2/manager_test.go
```diff
@@ -194,21 +199,35 @@ func TestKill(t *testing.T) {
 
 func TestMoveTo(t *testing.T) {
 	checkCgroupMode(t)
-	manager, err := NewManager(defaultCgroup2Path, "/test1", ToResources(&specs.LinuxResources{}))
+
+	manager, err := NewManager(defaultCgroup2Path, "/test-moveto-1", ToResources(&specs.LinuxResources{}))
```
manugupt1, line 203:
```suggestion
	manager, err := NewManager(defaultCgroup2Path, "/src-cgroup", ToResources(&specs.LinuxResources{}))
```

cgroup2/manager_test.go
manugupt1:
```suggestion
	destination, err := NewManager(defaultCgroup2Path, "/dst-cgroup", ToResources(&specs.LinuxResources{}))
```

manugupt1: Soemthing like these to make the test clearer?

dcantah: I like having the test prefix and something denoting what test, as if there is another leak like this it's pretty easy to deduce where it came from.

dcantah: Although test-moveto-dst and test-moveto-src are fine with me

manugupt1 (approved):

estesp (approved):

kzys (approved):

PR #283, opened 2023-04-07
cgroup2: rm/simplify some code

kolyshkin:
Please see individual commit messages for details.

This should make #381 way simpler

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -523,7 +523,7 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 	if err != nil {
 		return nil, err
 	}
-	out := make(map[string]interface{})
+	out := make(map[string]uint64)
 	for _, controller := range controllers {
 		switch controller {
 		case "cpu", "memory":
@@ -535,7 +535,7 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 			}
 		}
 	}
-	memoryEvents := make(map[string]interface{})
+	memoryEvents := make(map[string]uint64)
 	if err := readKVStatsFile(c.path, "memory.events", memoryEvents); err != nil {
 		if !os.IsNotExist(err) {
 			return nil, err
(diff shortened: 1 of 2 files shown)


dcantah (approved):

cgroup2/manager.go
```diff
@@ -541,14 +535,6 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
...
 		}
 	}
-	for _, name := range singleValueFiles {
-		if err := readSingleFile(c.path, name, out); err != nil {
-			if os.IsNotExist(err) {
-				continue
-			}
-			return nil, err
-		}
-	}
```
dcantah, line 551: I guess there is a slight distinction in that the old behavior, if we failed to read the file and it wasn't because it didn't exist we'd error out of Stat(), but.. not sure that's worth keeping around like 3 of the same function

kolyshkin: Yes, I saw that one and did not think it's a deal breaker ever.

cgroup2/manager.go
```diff
@@ -519,17 +518,12 @@ func (c *Manager) MoveTo(destination *Manager) error {
...
-	"pids.max",
-}
-
 func (c *Manager) Stat() (*stats.Metrics, error) {
 	controllers, err := c.Controllers()
 	if err != nil {
 		return nil, err
 	}
-	out := make(map[string]interface{})
+	out := make(map[string]uint64)
```
dcantah, line 526: Oh nice, looking now this is the better option as I don't think any of the stat files can have "max"

cgroup2/manager.go
dcantah: I don't either (you can resolve this, just wanted to point out the only behavioral difference I could see in the first commit)

dcantah (approved):

manugupt1 (approved):

dims (approved):
LGTM

(tried https://go.dev/play/p/eKHaT5V5RJ5)

PR #284, opened 2023-04-10
Cgroup1: Lower buffer size for uint reads

dcantah:
This was partly a brain fart on my end during review for this, we should only need 20 bytes to handle reading the largest unsigned 64 bit integer (21 if we for some reason wanted to read the line feed that's in all of these files to). Lets use 32 for a nice power of two though.

### DIFF
--- a/cgroup1/utils.go
+++ b/cgroup1/utils.go
@@ -137,7 +137,9 @@ func readUint(path string) (uint64, error) {
 	}
 	defer f.Close()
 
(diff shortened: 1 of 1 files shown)


manugupt1 (approved):

manugupt1:
We can close this: https://github.com/containerd/cgroups/pull/291 as I incorporated your changes here.

dcantah:
@manugupt1 perfect!

PR #285, opened 2023-04-11
Cgroup2: Testify all tests

dcantah:
Use require/assert for all the cgroup2 tests. These tests were in a weird spot where some of them used a mix of testify.assert, and some used `testify.require` in place of `if err != nil { t.Fatal() }`. This just moves everything over to `require.NoError` and friends.

### DIFF
--- a/cgroup2/manager_test.go
+++ b/cgroup2/manager_test.go
@@ -35,43 +35,35 @@ func TestEventChanCleanupOnCgroupRemoval(t *testing.T) {
 
 	cmd := exec.Command("cat")
 	stdin, err := cmd.StdinPipe()
-	if err != nil {
-		t.Fatalf("Failed to create cat process: %v", err)
-	}
-	if err := cmd.Start(); err != nil {
-		t.Fatalf("Failed to start cat process: %v", err)
-	}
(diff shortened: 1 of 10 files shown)


cgroup2/manager_test.go
```diff
@@ -147,35 +139,31 @@ func TestSystemdFullPath(t *testing.T) {
...
 	// Verify we have 5 pids before beginning Kill below.
 	pids, err := manager.Procs(true)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(pids) != 5 {
-		t.Fatalf("expected 5 pids, got %d", len(pids))
-	}
+	require.NoError(t, err)
+	require.Equal(t, numProcs, len(pids), "pid count unexpected")
```
samuelkarp, line 162: testify does have a `Len` method too

```suggestion
	require.Len(t, pids, numProcs, "pid count unexpected")
```

samuelkarp (approved):

cgroup2/manager_test.go
dcantah: ack, swapping to this!

samuelkarp (approved):

PR #286, opened 2023-04-11
Cgroup2: Cleanup procs in TestKill

dcantah:
If the test fails before we get to calling manager.Kill we'll leak the 5 sleep processes, and its harmless to call this
at the end of the test also. This additionally removes the cmd.Process nil check as the comment in the stdlib states that
cmd.Process will be filled in if Start returns successfully.

### DIFF
--- a/cgroup2/manager_test.go
+++ b/cgroup2/manager_test.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"syscall"
 	"testing"
 	"time"
 
@@ -147,9 +148,13 @@ func TestKill(t *testing.T) {
 	)
 	for i := 0; i < numProcs; i++ {
 		cmd := exec.Command("sleep", "infinity")
+		// Don't leak the process if we fail to join the cg,
+		// send sigkill after tests over.
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Pdeathsig: syscall.SIGKILL,
+		}
 		err = cmd.Start()
 		require.NoError(t, err)
-		require.NotNil(t, cmd.Process, "process is nil")
 
 		err = manager.AddProc(uint64(cmd.Process.Pid))
 		require.NoError(t, err)


manugupt1 (approved):

manugupt1:
Just a small comment: Should we kill all managers on cleanup using manager.Kill?

dcantah:
@manugupt1 I did a Kill on anything that actually launched any processes, I'd say it's case by case BUT we could add a test helper in a follow up here to check the number of processes running -> Call Kill if there is any -> Always call delete

manugupt1:
Oh okay! I was just wondering.

dcantah:
I think it's a good idea! It's easy to forget to cleanup so if future authors see we always just call 'CleanupCgroup' or something similar the situation would be better I think

cgroup2/manager_test.go
```diff
@@ -156,9 +156,10 @@ func TestKill(t *testing.T) {
 		if err := cmd.Start(); err != nil {
 			t.Fatal(err)
 		}
-		if cmd.Process == nil {
-			t.Fatal("Process is nil")
-		}
+		t.Cleanup(func() {
+			_ = cmd.Process.Kill()
```
fuweid, line 160: Maybe we should consider to add Pdeathsig to the process. Even if the test process has been killed and no chance to kill the sleep, the kernel will help us to do that.

REF: https://pkg.go.dev/syscall#SysProcAttr

dcantah: Not a bad idea, let me trial this tonight

dcantah: Swapped to this, thanks @fuweid

fuweid (approved):
LGTM

dcantah:
Oh woops, I just realized your comment for calling `manager.Kill` was on this PR and not https://github.com/containerd/cgroups/pull/282. So the rationale here is if the process failed to join the cg we'd leak it. There's another test that adds a process to a cg here (`TestMoveTo`) which is annoying as we can't call Kill on the cg, so I've changed that test to launch `sleep infinity` as well, I'll add the same Pdeathsig option on that PR.

dcantah:
Resolved merge conflicts from https://github.com/containerd/cgroups/commit/55c197e1d2d12f14193ffaac88166feaf4ddc932

estesp (approved):

PR #290, opened 2023-04-23
NewSystemd handles UnitExists when starting units

### LINKED ISSUE #279: NewSystemd must not ignore UnitExists
This is the same issue as described in https://github.com/opencontainers/runc/issues/3780

In short, `NewSystemd` ignores `UnitExists` error from systemd, as a result the PID is not added to the proper systemd unit and thus the cgroup.

See the above issue for details as to how that may happen.

See the https://github.com/opencontainers/runc/pull/3782 as to how this might be fixed.

mmerkes:
As done in this runc PR
https://github.com/opencontainers/runc/pull/3782,
UnitExists errors are no longer ignored when starting units. If it already exists, we attempt to reset the failed unit and retry once. Otherwise, we fail. See https://github.com/containerd/cgroups/issues/279 for more context.

Fixes #279 

In generally, [I'm following runc's example](https://github.com/kolyshkin/runc/blob/82bc89cd10eb5fa5fdb5d77e41b7ea3c07f9d834/libcontainer/cgroups/systemd/common.go#L127-L174) and making this code more robust. See code and commit for more details. Ideally, we'd reuse the code, but don't currently have a dependency on `runc` as a library here and I don't have enough context to know if that would be reasonable or not.

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -870,20 +870,67 @@ func NewSystemd(slice, group string, pid int, resources *Resources) (*Manager, e
 			newSystemdProperty("TasksMax", uint64(resources.Pids.Max)))
 	}
 
-	statusChan := make(chan string, 1)
-	if _, err := conn.StartTransientUnitContext(ctx, group, "replace", properties, statusChan); err == nil {
-		select {
-		case <-statusChan:
-		case <-time.After(time.Second):
-			logrus.Warnf("Timed out while waiting for StartTransientUnit(%s) completion signal from dbus. Continuing...", group)
-		}
-	} else if !isUnitExists(err) {
+	if err := startUnit(conn, group, properties, pid == -1); err != nil {
 		return &Manager{}, err
 	}
 
 	return &Manager{
 		path: path,
 	}, nil
+}
+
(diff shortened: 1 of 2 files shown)


mmerkes:
@kolyshkin Let me know if you have feedback here since you opened the original issue and made the runc change. Tx!

kolyshkin (approved):
Alas, I am not a maintainer and can't say which uses/users this method has.

But nevertheless a fix needs to be made, because the current behavior (of not adding the container into a proper cgroup) is totally unacceptable (and can, in some circumstances, be qualified as a security issue).

So, I'm adding my non-binding LGTM, but this needs a review from the actual maintainers of this package.

AkihiroSuda (changes requested):
Please squash commits

mmerkes:
> Please squash commits

Updated. Falsely assumed that the merge would squash. @AkihiroSuda

AkihiroSuda (approved):

thaJeztah (approved):
LGTM

mmerkes:
@AkihiroSuda This good to be merged?

PR #291, opened 2023-04-28
Improve performance by for pid stats (cgroups1) re-using readuint

manugupt1:
cc @kolyshkin @dcantah 

I re-used the optimized readuint in cgroups to improve performance. This should not change the behavior for max as readuint returns 0 and when a file is not present; an error is still returned.

PTAL

Benchstat results:

```
➜  cgroups git:(main) ✗ benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/containerd/cgroups/v3/cgroup1
           │   old.txt   │              new.txt               │
           │   sec/op    │   sec/op     vs base               │
TestPids-8   19.32µ ± 4%   17.39µ ± 2%  -9.97% (p=0.000 n=10)

           │   old.txt   │              new.txt               │
           │    B/op     │    B/op     vs base                │
TestPids-8   1344.0 ± 0%   624.0 ± 0%  -53.57% (p=0.000 n=10)

           │  old.txt   │              new.txt               │
           │ allocs/op  │ allocs/op   vs base                │
TestPids-8   13.00 ± 0%   11.00 ± 0%  -15.38% (p=0.000 n=10)
➜  cgroups git:(main) ✗
```

### DIFF
--- a/cgroup1/pids.go
+++ b/cgroup1/pids.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
-	"strings"
 
 	v1 "github.com/containerd/cgroups/v3/cgroup1/stats"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
@@ -67,15 +66,9 @@ func (p *pidsController) Stat(path string, stats *v1.Metrics) error {
 	if err != nil {
 		return err
 	}
-	var max uint64
-	maxData, err := os.ReadFile(filepath.Join(p.Path(path), "pids.max"))
+	max, err := readUint(filepath.Join(p.Path(path), "pids.max"))
 	if err != nil {
 		return err
-	}
-	if maxS := strings.TrimSpace(string(maxData)); maxS != "max" {
-		if max, err = parseUint(maxS, 10, 64); err != nil {
-			return err
-		}
 	}
 	stats.Pids = &v1.PidsStat{
 		Current: current,
--- a/cgroup1/pids_test.go
+++ b/cgroup1/pids_test.go
@@ -186,3 +186,50 @@ func TestPidsOverflowMax(t *testing.T) {
 		t.Fatal("expected not nil err")
 	}
 }
+
+func BenchmarkTestPids(b *testing.B) {
+
+	mock, err := newMock(b)
+	if err != nil {
+		b.Fatal(err)
+	}
+	defer func() {
+		if err := mock.delete(); err != nil {
+			b.Errorf("failed delete: %v", err)
+		}
+	}()
+
+	pids := NewPids(mock.root)
+	if pids == nil {
+		b.Fatal("pids is nil")
+	}
+	resources := specs.LinuxResources{
+		Pids: &specs.LinuxPids{
+			Limit: 10,
+		},
+	}
+
+	err = pids.Create("test", &resources)
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	current := filepath.Join(mock.root, "pids", "test", "pids.current")
+	if err = os.WriteFile(
+		current,
+		[]byte(strconv.Itoa(5)),
+		defaultFilePerm,
+	); err != nil {
+		b.Fatal(err)
+	}
+
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		metrics := v1.Metrics{}
+		err = pids.Stat("test", &metrics)
+		if err != nil {
+			b.Fatal(err)
+		}
+	}
+}
--- a/cgroup1/utils.go
+++ b/cgroup1/utils.go
@@ -18,6 +18,7 @@ package cgroup1
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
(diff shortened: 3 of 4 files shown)


cgroup1/utils.go
```diff
@@ -137,7 +137,7 @@ func readUint(path string) (uint64, error) {
 	}
 	defer f.Close()
 
-	b := make([]byte, 128) // Chose 128 as some files have leading/trailing whitespaces and alignment
+	b := make([]byte, 32) // Chose 32 as some files have leading/trailing whitespaces and alignment
```
dcantah, line 140: I'd just plop the comment on my old pr here. It seems for every cgroup file there's just a single line feed at the end and not trailing whitespace, and we expect to be able to be able to read a uint64 in 20 bytes etc.

cgroup1/pids_test.go
```diff
@@ -186,3 +186,47 @@ func TestPidsOverflowMax(t *testing.T) {
...
+		}
+	}()
+
+	pids := NewPids(mock.root)
+	if pids == nil {
+		b.Fatal("pids is nil")
+	}
+	resources := specs.LinuxResources{}
+	resources.Pids = &specs.LinuxPids{}
+	resources.Pids.Limit = int64(10)
```
dcantah, line 208:
```suggestion
	resources := specs.LinuxResources{
	    Pids: &specs.LinuxPids{
	         Limit: 10,
	    },
	}
	
```

cgroup1/pids.go
```diff
@@ -67,16 +66,10 @@ func (p *pidsController) Stat(path string, stats *v1.Metrics) error {
 	if err != nil {
 		return err
 	}
-	var max uint64
-	maxData, err := os.ReadFile(filepath.Join(p.Path(path), "pids.max"))
+	max, err := readUint(filepath.Join(p.Path(path), "pids.max"))
```
dcantah, line 69: Isn't the behavior different here? `readUint` should fail if "max" exists in the file as it only knows how to parse integers, so we'd bail on line 71. Whereas before we'd carry on to filling in the stats, granted the `max` variable will be left as 0 it seems if "max" was in pids.max, but point being this method would still succeed

manugupt1: Right; I changed readUint to return MaxUint64 now. The behavior is still different but it is consistent with cgroups2. I think this change should be fine now.

dcantah: Yea it is odd to me how we returned 0 here before.. would need to see if any users relied on this behavior with a github search maybe..

dcantah: (I think returning MaxUint64 makes sense)

manugupt1: I think it makes sense at least for v3; v1 is imported by so many packages that I am not confident.

1. https://pkg.go.dev/github.com/containerd/cgroups?tab=importedby
2. https://pkg.go.dev/github.com/containerd/cgroups/v3@v3.0.1?tab=importedby


Thinking of reverting back to 0 and adding a comment to say that we preserve it for backward compatibility.
What do you think?

dcantah: I'd err on the side of staying safe, so your reasoning seems sane and I agree (also sorry for the radio silence here 🫠)

dcantah: I suppose 0 could be interpreted as no limit, so that might've been the rationale

dcantah (approved):

cgroup1/pids.go
mmerkes: Zero wouldn't be a valid value anyway as zero, so that's a fair rationale.

cgroup1/utils.go
```diff
@@ -137,13 +138,19 @@ func readUint(path string) (uint64, error) {
 	}
 	defer f.Close()
 
-	b := make([]byte, 128) // Chose 128 as some files have leading/trailing whitespaces and alignment
+	// We should only need 20 bytes for the max uint64, but for a nice power of 2
+	// lets use 32.
```
mmerkes, line 142: Per the original comment, is it realistic that this might have leading/trailing whitespace? Seems unlikely, but anyone who does would start breaking if they had a lot of it.

dcantah: It seems the files simply have one line feed.

mmerkes: I guess trailing whitespace isn't actually an issue because it will only read `len(b)` bytes and stop anyway, so the only way to break this would be to have a bunch of leading whitespace, which seems unlikely in this case.

dcantah: Right, it's mostly irrelevant if we only read a fixed size of bytes that will always be larger than a uint64's length. The string gets trimmed and all that's left is what we care about.

cgroup1/pids.go
manugupt1: @dcantah No worries! thanks for the review! this was a really fun exercise :)
thanks @mmerkes for the review and helping building understanding

kzys (approved):

AkihiroSuda (approved):

PR #296, opened 2023-06-10
cgroup1 delete: proceed to the next subsystem when a cgroup is not found

singholt:
Signed-off-by: Anuj Singh <singholt@amazon.com>

The cgroup1 Delete method checks for running processes within each subsystem, before deleting the cgroup from it. 

I noticed that there are cases when removing a cgroup from one subsystem, also removes it from others.
For example, on Amazon Linux (kernel 4.14), removing a cgroup from `net_cls` subsystem also removes the cgroup from `net_prio`. This is because both `net_cls` and `net_prio` have a symlink to a common dir `net_cls, net_prio`. Similarly, removing a cgroup from `cpu` also removes it from the `cpuacct` subsystem, because of a symlink to `cpu,cpuacct`.

```
ls -l /sys/fs/cgroup/net_cls /sys/fs/cgroup/net_prio
/sys/fs/cgroup/net_cls -> net_cls,net_prio
/sys/fs/cgroup/net_prio -> net_cls,net_prio 
```

In such a case the Delete() errors out with an error like 

```
lstat /sys/fs/cgroup/net_prio/test: no such file or directory
```

Sample code: https://gist.github.com/singholt/2a600802c0f74c8acdacd247760dc291 that fails on the Delete() without this fix but passes with it.

This change handles the above case, when a query to get cgroup processes within a subsystem returns a `fs.ErrNotExist`. In such a case, we move to the next subsystem, instead of erroring out of the Delete() immediately. This ensures that the cgroup is deleted from all subsystems that have it.

As per cgroups v1 [documentation](https://man7.org/linux/man-pages/man7/cgroups.7.html#CGROUPS_VERSION_1), since files in a cgroup directory cannot and need not be removed, it is safe to assume that if `cgroup.procs` doesn't exist for a cgroup (within a subsystem), the cgroup has been removed from the subsystem.

This PR has two commits - first addressing the problem above, with the unit test updated and second, just fixing a typo.

### DIFF
--- a/cgroup1/cgroup.go
+++ b/cgroup1/cgroup.go
@@ -41,7 +41,7 @@ func New(path Path, resources *specs.LinuxResources, opts ...InitOpts) (Cgroup, 
 			return nil, err
 		}
 	}
-	subsystems, err := config.hiearchy()
+	subsystems, err := config.hierarchy()
 	if err != nil {
 		return nil, err
 	}
@@ -79,7 +79,7 @@ func Load(path Path, opts ...InitOpts) (Cgroup, error) {
 		}
 	}
 	var activeSubsystems []Subsystem
-	subsystems, err := config.hiearchy()
+	subsystems, err := config.hierarchy()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


estesp (approved):
LGTM

singholt:
thanks for reviewing Phil! It looks like I may need another review, perhaps @kzys ? I'd appreciate if you could take a look 😄

dcantah (approved):

PR #298, opened 2023-06-16
CI on macos-11 with the goal to not fail even if it's not supported

bufferoverflow:
related to https://gitlab.com/gitlab-org/gitaly/-/merge_requests/5547#note_1428633915

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
       matrix:
         go-version: [1.19.x, 1.20.x]
         # Ubuntu-20.04 has cgroups v1 default; Ubuntu-22.04 has cgroups v2 default.
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-20.04, ubuntu-22.04, macos-11]
 
     runs-on: ${{ matrix.os }}
     steps:
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ all: cgutil
 	go build -v
 
 cgutil:
-	cd cmd/cgctl && go build -v
+	cd cmd/cgctl && GOOS=linux go build -v
 
 proto:
 	protobuild --quiet ${PACKAGES}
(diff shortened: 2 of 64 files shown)


kzys:
Hmm, should gitaly have a no-op implementation? Make containerd/cgroups "work" for macOS doesn't make much sense, since almost all features don't work there.

bufferoverflow:
> Hmm, should gitaly have a no-op implementation? Make containerd/cgroups "work" for macOS doesn't make much sense, since almost all features don't work there.

Yes, I'm also a bit unsure. But somehow I think even if `cgroups` are not available on some or as per today all the non-Linux operating systems it's worth to define that precise within all the source files and provide a useful value such as [`Mode()`](https://github.com/containerd/cgroups/pull/298/files#diff-a264c0eb0c1aa3e25fb6a004557600e52397376f6b76764604831094f1ba4393R21). Especially to simplify peoples life who are facing a complex piece of software with tons of dependencies.

If the target os is not defined people will build mocks or no-op stubs all over the planet, why not solving this once upstream?

adlternative:
Yes, if the cgroup upstream could consider non-Linux systems, functions like `Mode()` would be easier to use.

estesp:
I think the core issue here is that this is a subproject exists specifically to support cgroups, which are only a Linux kernel concept today. Making the project "viable" for non-Linux has no value as there is no way to use this Linux feature on macOS natively. Our main project, containerd, imports and uses this project on multiple platforms, including macOS, without us ever having to stub the entire set of functions in this subproject, because that codebase doesn't try to use these features at all on macOS, for example. 

> If the target os is not defined people will build mocks or no-op stubs all over the planet, why not solving this once upstream?

I would suggest that based on the # of projects taking a dependency on this sub-project of containerd, it has not been extremely difficult or time consuming to write software that simply doesn't even try to use cgroups capabilities when the OS is not Linux. If that isn't the case with the project you are trying to make viable on macOS, maybe one of us could take a look and help in some way?

bufferoverflow:
@estesp You are right. I have no mac and so I did a short session with @max-wittig and we made it work within a short time on mac as well. Sorry for the noise here and have a great weekend!

PR #302, opened 2023-08-04
Submit/systemd optional

metux:


dcantah:
Looks like this needs a rebase, sorry for the delay here..

PR #306, opened 2023-11-10
go.mod: Upgrade github.com/cilium/ebpf v0.12.3

AlbertoBarba:
Needed to fix a breaking change introduced in `golang.org/x/sys`
https://github.com/golang/go/issues/63969


mavimo (approved):

AlbertoBarba:
It seems the breaking change is going to be reverted in `golang.org/x/sys@v0.15.0`. It’s probably better to wait till it gets released

Issue #307, opened 2023-11-22
oom killer cannot be re-enabled


kestrelcjx:
https://github.com/containerd/cgroups/blob/a0ae1c2a2c40df548614588dd079a4a790971918/cgroup1/memory.go#L454

kestrelcjx:
https://github.com/containerd/cgroups/blob/a0ae1c2a2c40df548614588dd079a4a790971918/cgroup1/memory.go#L453-L459

dcantah:
Fixed via https://github.com/containerd/cgroups/pull/311

PR #311, opened 2023-11-28
fix(): support re-enabling oom killer refs #307

kestrelcjx:
DisableOOMKiller will be ignored if it is false, which will result in oomkiller being unable to be re-enabled after this setting is disabled.

### DIFF
--- a/cgroup1/memory.go
+++ b/cgroup1/memory.go
@@ -454,6 +454,9 @@ func getOomControlValue(mem *specs.LinuxMemory) *int64 {
 	if mem.DisableOOMKiller != nil && *mem.DisableOOMKiller {
 		i := int64(1)
 		return &i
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

cgroup1/memory.go
```diff
@@ -454,6 +454,9 @@ func getOomControlValue(mem *specs.LinuxMemory) *int64 {
 	if mem.DisableOOMKiller != nil && *mem.DisableOOMKiller {
 		i := int64(1)
 		return &i
+	} else if mem.DisableOOMKiller != nil && !*mem.DisableOOMKiller {
```
AkihiroSuda, line 457: No need to repeat `if mem.DisableOOMKiller != nil` twice

PR #312, opened 2023-11-30
feat: add Threads for cgroupv2

Rouzip:
Add `Threads` for cgroup v2 like `Tasks(Name, bool) ([]Task, error)` in cgroup v1.

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"math"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


Rouzip:
PTAL, thanks. @AkihiroSuda

cgroup2/manager.go
```diff
@@ -504,6 +504,32 @@ func (c *Manager) Procs(recursive bool) ([]uint64, error) {
...
+		if name != cgroupThreads {
+			return nil
+		}
+		procs, err := parseCgroupProcsFile(p)
+		if err != nil {
+			return err
+		}
+		threads = append(threads, procs...)
+		return nil
+	})
```
dcantah, line 529: Given this is shared between `Procs` and this, can we split this into a helper that just takes in the name to care about? e.g. pass in cgroupThreads for this and cgroupProcs for `Procs`

Rouzip: Hello, if you mean something like [this](https://github.com/Rouzip/runc/blob/50e7b9cb9605aa3409f96267aa341b15f24f862d/libcontainer/cgroups/utils.go#L140-L162)?

Rouzip:
Hello, any advices are welcome. PTAL @dcantah

dcantah (approved):

AkihiroSuda (approved):

Issue #309, opened 2023-11-28
ToResources method has a bug in the conversion of swap memory limits.


kestrelcjx:
https://github.com/containerd/cgroups/blob/a0ae1c2a2c40df548614588dd079a4a790971918/cgroup2/utils.go#L155-L187

kestrelcjx:
<img width="1167" alt="截屏2023-11-28 上午11 11 58" src="https://github.com/containerd/cgroups/assets/38725513/28b4f3bc-3ace-40b2-a13b-d9968bc42e54">

kestrelcjx:
https://github.com/containerd/cgroups/blob/a0ae1c2a2c40df548614588dd079a4a790971918/cgroup2/utils.go#L178C2-L178C2
resources.Memory.Swap may need to be changed to swap-max

dcantah:
Fixed via https://github.com/containerd/cgroups/pull/314

PR #314, opened 2023-12-02
Linux.Swap is defined as memory+swap combined, while in cgroup v2 swap is a separate value

zouyee:
fix： https://github.com/containerd/cgroups/issues/309


zouyee:
@AkihiroSuda PTAL

AkihiroSuda (approved):
Thanks

PR #318, opened 2023-12-31
cg2: Don't read cgroup.procs when deleting threaded cg

dcantah:
Reading cgroup.procs seems to return ENOTSUPP when threaded, so check the type of the cg when going to delete and read the relevant file.

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -468,13 +468,33 @@ func (c *Manager) fallbackKill() error {
 }
 
 func (c *Manager) Delete() error {
-	// kernel prevents cgroups with running process from being removed, check the tree is empty
-	processes, err := c.Procs(true)
+	var (
+		tasks    []uint64
+		threaded bool
+	)
+	// Kernel prevents cgroups with running process from being removed,
+	// check the tree is empty.
+	//
+	// Pick the right file to read based on the cgs type.
+	cgType, err := c.GetType()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


cgroup2/manager.go
```diff
@@ -468,13 +468,33 @@ func (c *Manager) fallbackKill() error {
...
+	} else {
+		tasks, err = c.Procs(true)
+	}
 	if err != nil {
 		return err
 	}
-	if len(processes) > 0 {
-		return fmt.Errorf("cgroups: unable to remove path %q: still contains running processes", c.path)
+	if len(tasks) > 0 {
+		return fmt.Errorf("cgroups: unable to remove path %q: still contains running tasks", c.path)
```
cpuguy83, line 497: Curious why we don't try to remove and check for `EBSUY` instead?

dcantah: I think (and this is the same for what was here today) this is mainly for a nice error, so I tried to keep what was here today. I suppose we could just immediately `remove(c.path)` (which tries to handle a reasonable amount of ebusy looping) although afterwards we'd still need to know what file to read to surface some decent error if the remove didn't go through.

dmcgowan (approved):

cpuguy83 (approved):

dcantah:
@dmcgowan This should be ready to check in, didn't realize I got two 😆

fuweid (approved):

PR #322, opened 2024-01-04
Add `EOPNOTSUPP` to err filter for PSI data

mahendrapaipuri:
When PSI is not enabled, the file reads will end up in `Operation not supported` error. Here is the relevant strace

```
mmap(NULL, 139264, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x154ea7496000
read(3, 0x154ea7497000, 131072)         = -1 EOPNOTSUPP (Operation not supported)
```

This PR adds `EOPNOTSUPP` to filter as well besides `ENOTSUP`.

### DIFF
--- a/cgroup2/utils.go
+++ b/cgroup2/utils.go
@@ -514,7 +514,7 @@ func getStatPSIFromFile(path string) *stats.PSIStats {
 	}
 
 	if err := sc.Err(); err != nil {
-		if !errors.Is(err, unix.ENOTSUP) {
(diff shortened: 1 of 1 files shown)


cpuguy83 (approved):

dcantah (approved):

klihub:
@mahendrapaipuri It would be preferable to squash those commits into a single one.

mahendrapaipuri:
@klihub Makes sense! Will do it now!

dmcgowan (approved):

PR #323, opened 2024-01-05
support cpu idle and burst for cgroupv2

zouyee:


cpuguy83:
I'm having trouble finding the kernel docs on these. Have any pointers?

zouyee:
> I'm having trouble finding the kernel docs on these. Have any pointers?

[sched: Cgroup SCHED_IDLE support](https://lore.kernel.org/lkml/162971078674.25758.15464079371945307825.tip-bot2@tip-bot2/#r) is introduced in Linux 5.15
- https://lore.kernel.org/lkml/162971078674.25758.15464079371945307825.tip-bot2@tip-bot2/#r

Burstable CFS controller is landed in Linux 5.14 
https://github.com/torvalds/linux/commit/f4183717b370ad28dd0c0d74760142b20e6e7931#diff-cc1a82129952a910fdc4292448c2a097a2ba538bebefcf3c06381e45639ae73e

zouyee:
@dcantah @mxpv PTAL

cgroup2/manager.go
```diff
@@ -849,9 +849,16 @@ func NewSystemd(slice, group string, pid int, resources *Resources) (*Manager, e
 			newSystemdProperty("MemoryMax", uint64(*resources.Memory.Max)))
 	}
 
-	if resources.CPU != nil && resources.CPU.Weight != nil && *resources.CPU.Weight != 0 {
-		properties = append(properties,
-			newSystemdProperty("CPUWeight", *resources.CPU.Weight))
+	if resources.CPU != nil {
+		// Do not add duplicate CPUWeight property
+		if resources.CPU.Idle != nil && *resources.CPU.Weight != 0 {
```
cpuguy83, line 854: Can you add some extra comments on this new case?

cpuguy83: Also I think you need a nil check for `CPU.Weight`?

cpuguy83 (changes requested):

cgroup2/manager.go
```diff
@@ -869,9 +869,20 @@ func NewSystemd(slice, group string, pid int, resources *Resources) (*Manager, e
...
-	if resources.CPU != nil && resources.CPU.Weight != nil && *resources.CPU.Weight != 0 {
-		properties = append(properties,
-			newSystemdProperty("CPUWeight", *resources.CPU.Weight))
+	if resources.CPU != nil {
+		// Systemd v252 added support for setting cgroup v2 cpu.idle in systemd/systemd#23299
+		// The way it works is
+		// if CPUWeight == 0, cpu.idle is set to 1;
+		// if CPUWeight != 0, cpu.idle is set to 0.
+		// Do not add duplicate CPUWeight property
+		if resources.CPU.Idle != nil && resources.CPU.Weight != nil && *resources.CPU.Weight != 0 {
```
cpuguy83, line 878: I'm not following this logic.
Right now if `resources.CPU.Idle` is set and `resources.CPU.Weight` is set to a non-zero value then we end up with `CPUWeight` being set twice.

PR #335, opened 2024-04-21
Add support for io.latency interface file in IO controller

Xeway:
This feature can be very useful in some situation.

Sources that explains how it works, how to use it, etc.:
- https://facebookmicrosites.github.io/cgroup2/docs/io-controller.html#protecting-workloads-with-iolatency
- https://www.kernel.org/doc/html/v5.10/admin-guide/cgroup-v2.html#io-latency

NB: on some Linux distributions (like Debian-based ones), this feature is not enabled, probably because:

> Note, this is an experimental interface and could be changed someday.

<sup>[Source](https://cateee.net/lkddb/web-lkddb/BLK_CGROUP_IOLATENCY.html)</sup>


Xeway:
Force pushed because DCO was missing for each commit

PR #342, opened 2024-06-21
go.mod: update to go1.19 as minimum, and add go1.22.x to GHA

thaJeztah:
### go.mod: update to go1.19 as minimum

This is the oldest version tested in CI, so set this as the
minimum version.

### gha: also test on go1.22.x

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.20.x'
+          go-version: '1.22.x'
           cache-dependency-path: src/github.com/containerd/cgroups
 
       - name: Project checks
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
     strategy:
       matrix:
-        go-version: [1.19.x, 1.20.x]
+        go-version: [1.19.x, 1.21.x, 1.22.x]
```
AkihiroSuda, line 53: We can drop 1.19

AkihiroSuda (approved):

.github/workflows/ci.yml
thaJeztah: Yeah, my thinking here was to keep a minimum version (unless we had a strong reason to drop it), then check current versions.

I'm fine with dropping 1.19 by now probably, although "if it works", then 🤷 .

I'll drop the 1.20, and keep the upper-bound to 1.21, 1.22 as suggested.

thaJeztah: oh! I read the diff wrong, thought it was a suggestion, but I already changed to 1.21, 1.22 🤦 😂

dmcgowan (approved):

PR #344, opened 2024-06-23
Allow for passing properties to NewSystemd

shell-skrimp:
I had a need to disable some functionality but couldnt do so with current implementation so I made a tweak that allows for passing custom properties.

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -830,7 +830,7 @@ func dashesToPath(in string) string {
 	return path
 }
 
-func NewSystemd(slice, group string, pid int, resources *Resources) (*Manager, error) {
+func NewSystemd(slice, group string, pid int, resources *Resources, properties []systemdDbus.Property) (*Manager, error) {
 	if slice == "" {
 		slice = defaultSlice
 	}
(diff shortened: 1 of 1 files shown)


dcantah:
@shell-skrimp Unfortunately this changes the public API of `NewSystemd` which would be a breaking change. I think we could go one of two ways if we want this feature:
1. Introduce a new "constructor" that takes in the properties. Basically what you have right now just with a new name like `NewSystemdWithProperties`
2. Add a variadic options parameter as the last parameter. Ex:
```go
func NewSystemd(slice, group string, pid int, resources *Resources, props ...SystemdProperties) (*Manager, error) {

// or the below, where the only option we have at the moment is just supplying a slice of properties.
func NewSystemd(slice, group string, pid int, resources *Resources, opts ...SystemdOptions) (*Manager, error) {
```

dcantah:
@shell-skrimp Unfortunately this changes the public API of NewSystemd which would be a breaking change. You could introduce a new "constructor" that takes in the properties. Basically what you have right now just with a new name like NewSystemdWithProperties

samuelkarp:
> I had a need to disable some functionality

Can you tell us more about your use-case?  As @dcantah noted this is a breaking change; a new constructor would be a better option but as of now I don't think we know of a particular use-case for this in containerd.

PR #348, opened 2024-09-03
Read cpu.stat regardless if controller enabled.

### LINKED ISSUE #347: Support cgroup v2 CPU stats when controller not enabled
`manager.Stat()` iterates over enabled controllers to construct statistics, namely for memory and CPU. If a controller is not enabled, the statistics are not provided. 

https://github.com/containerd/cgroups/blob/0c03de4a3d82a5f02f455ccc8174cb0dc9c2a532/cgroup2/manager.go#L565-L575

However, regardless of whether the `cpu` controller is enabled,  cgroups v2 always provides the following statistics in `cpu.stat` :
- `usage_usec`
- `user_usec`
- `system_usec`

Systemd no longer enables this controller when the property `CPUAccounting=True`, meaning these statistics are not provided when they probably should be.  `Stat()` should be modified to fix this. 

 See https://docs.kernel.org/admin-guide/cgroup-v2.html#cpu-interface-files, systemd/systemd#10507

jay-mckay:
The unified hierarchy provides the cpu.stat file for every cgroup, regardless if the CPU controller is enabled (in fact, setting the systemd property CPUAccounting=True does not enable this controller because of this fact). It provides the usage_usec, user_usec, and system_usec by default. Instead of reading the stat for each enabled controller (CPU and memory), just attempt to read them each time the Stat() function is called.

Attempting to read the memory.stat file even if memory accounting is not enabled seems insignificant (some other files always have a read attempt, such as memory.current), and eliminates finding and looping over enabled controllers.

resolves #347

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -559,112 +559,132 @@ func (c *Manager) MoveTo(destination *Manager) error {
 }
 
 func (c *Manager) Stat() (*stats.Metrics, error) {
-	controllers, err := c.Controllers()
-	if err != nil {
-		return nil, err
-	}
-	// Sizing this avoids an allocation to increase the map at runtime;
-	// currently the default bucket size is 8 and we put 40+ elements
-	// in it so we'd always end up allocating.
-	out := make(map[string]uint64, 50)
-	for _, controller := range controllers {
-		switch controller {
-		case "cpu", "memory":
-			if err := readKVStatsFile(c.path, controller+".stat", out); err != nil {
-				if os.IsNotExist(err) {
-					continue
-				}
-				return nil, err
-			}
-		}
-	}
-	memoryEvents := make(map[string]uint64)
-	if err := readKVStatsFile(c.path, "memory.events", memoryEvents); err != nil {
-		if !os.IsNotExist(err) {
-			return nil, err
-		}
-	}
-
 	var metrics stats.Metrics
+	var err error
+
 	metrics.Pids = &stats.PidsStat{
 		Current: getStatFileContentUint64(filepath.Join(c.path, "pids.current")),
 		Limit:   getStatFileContentUint64(filepath.Join(c.path, "pids.max")),
 	}
-	metrics.CPU = &stats.CPUStat{
-		UsageUsec:     out["usage_usec"],
-		UserUsec:      out["user_usec"],
-		SystemUsec:    out["system_usec"],
-		NrPeriods:     out["nr_periods"],
-		NrThrottled:   out["nr_throttled"],
-		ThrottledUsec: out["throttled_usec"],
-		PSI:           getStatPSIFromFile(filepath.Join(c.path, "cpu.pressure")),
+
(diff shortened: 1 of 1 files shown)


cgroup2/manager.go
```diff
@@ -634,6 +617,13 @@ func (c *Manager) Stat() (*stats.Metrics, error) {
 		SwapMaxUsage:          getStatFileContentUint64(filepath.Join(c.path, "memory.swap.peak")),
 		PSI:                   getStatPSIFromFile(filepath.Join(c.path, "memory.pressure")),
 	}
+
+	memoryEvents := make(map[string]uint64)
+	if err := readKVStatsFile(c.path, "memory.stat", memoryStat); err != nil {
```
dcantah, line 622:
```suggestion
	if err := readKVStatsFile(c.path, "memory.events", memoryStat); err != nil {
```

dcantah:
Change looks good after the one comment is fixed

thaJeztah:
I see this PR branch also has some merge commits in it; we generally use a rebase flow to keep a clean commit history; when updating, could you do a rebase and squash the commits, so that there's only a single commit in the PR?

Let me know if you need help / instructions doing so!

jay-mckay:
Fix added and rebase complete, should only have one commit now. Thanks.

dcantah (approved):

cgroup2/manager.go
```diff
@@ -550,78 +550,61 @@ func (c *Manager) MoveTo(destination *Manager) error {
...
+		NrPeriods:     cpuStat["nr_periods"],
+		NrThrottled:   cpuStat["nr_throttled"],
+		ThrottledUsec: cpuStat["throttled_usec"],
 		PSI:           getStatPSIFromFile(filepath.Join(c.path, "cpu.pressure")),
 	}
+	memoryStat := make(map[string]uint64, 40)
+	if err := readKVStatsFile(c.path, "memory.stat", memoryStat); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, err
+		}
```
thaJeztah, line 574: So, while looking at this PR (and I think some of this is an existing issue) I wonder if it would make sense to skip assigning values in situations where the file does _not_ exist, or if we should skip all of that if it's not there, and explicitly construct an empty `&stats.MemoryStat{}`.

muzea:
This pull request has been very helpful to me. Hope it can be merged soon.

thaJeztah:
@jay-mckay let me know what you think about my suggestions in https://github.com/jay-mckay/cgroups/pull/1 - if you merge that PR, that commit should show up in this PR (if that's looking good to @dcantah , we can squash the commits after that)

jay-mckay:
@thaJeztah The changes look good to me, I have merged them in.

AkihiroSuda:
Needs rebase

jay-mckay:
@AkihiroSuda rebase complete and commits squashed.

AkihiroSuda (approved):

Issue #351, opened 2024-11-18
CI: lint is failing


AkihiroSuda:
https://github.com/containerd/cgroups/actions/runs/11691328166/job/32558246281

```
Error: src/github.com/containerd/cgroups/cgroup1/cgroup.go:37:32: undefined: specs (typecheck)
  func New(path Path, resources *specs.LinuxResources, opts ...InitOpts) (Cgroup, error) {
                                 ^
  Error: src/github.com/containerd/cgroups/cgroup1/cgroup.go:132:46: undefined: specs (typecheck)
  func (c *cgroup) New(name string, resources *specs.LinuxResources) (Cgroup, error) {
                                               ^
  Error: src/github.com/containerd/cgroups/cgroup1/cgroup.go:353:36: undefined: specs (typecheck)
  func (c *cgroup) Update(resources *specs.LinuxResources) error {
                                     ^
  Error: src/github.com/containerd/cgroups/cgroup1/systemd.go:150:10: undefined: dbus (typecheck)
  		Value: dbus.MakeVariant(units),
  		       ^
  Error: src/github.com/containerd/cgroups/cgroup1/utils.go:200:15: undefined: cgroups (typecheck)
  	x, _, err := cgroups.ParseCgroupFileUnified(path)
  	             ^
  Error: src/github.com/containerd/cgroups/cgroup1/utils.go:209:9: undefined: cgroups (typecheck)
  	return cgroups.ParseCgroupFileUnified(path)
  	       ^
  Error: src/github.com/containerd/cgroups/cgroup1/paths_test.go:108:25: undefined: cgroups (typecheck)
  	paths, unified, err := cgroups.ParseCgroupFromReaderUnified(r)
  	                       ^
  Error: src/github.com/containerd/cgroups/cgroup2/manager.go:1009:10: undefined: dbus (typecheck)
  		Value: dbus.MakeVariant(units),
  		       ^
  Error: src/github.com/containerd/cgroups/cgroup2/utils.go:384:28: undefined: dbus (typecheck)
  		if dbusError, ok := err.(dbus.Error); ok {
[...]
```

AlbertoBarba:
@AkihiroSuda i wasn't able to reproduce the linter output. However i've found out that the lint jobs where using the wrong go version due to a typo. On the PR the lint jobs are currently green

PR #352, opened 2024-11-19
fix: use the specified go version in lint jobs

AlbertoBarba:
Ref #351

Due to a typo in the workflow the lint jobs where installing the latest version of go, rather then the one specified.


AkihiroSuda (approved):
Thanks

PR #354, opened 2024-11-27
add inotify error when during init

akhilerm:
add error returned from `InotifyInit` along with custom error message.


akhilerm:
Closed for #355

PR #356, opened 2024-12-05
feat: add cpu max burst support

lucas-jacques:


cgroup2/cpu.go
AkihiroSuda, line 1: Please squash the commits

PR #324, opened 2024-01-27
Added cgroup type

yaoyinnan:
Add domain threaded and domain invalid cgroup type.

PR #358, opened 2024-12-10
cgroup2: Manager.Delete: handle both "threaded" and "domain threaded"

thaJeztah:
- relates to https://github.com/containerd/cgroups/pull/318
- relates to https://github.com/containerd/cgroups/pull/324

commit 6f5001dbae8ec248caad9b86c07c869f10ec5726 added special handling for threaded cgroup types. A later contribution added detection for "domain threaded" as known type, but did not update the handling to detect this type.

From the original PR;

> Reading cgroup.procs seems to return ENOTSUPP when threaded, so check
> the type of the cg when going to delete and read the relevant file.

An alternative could be to check both variants unconditionally, and to error if either Manager.Threads or Manager.Procs is non-zero.

### DIFF
--- a/cgroup2/manager.go
+++ b/cgroup2/manager.go
@@ -472,10 +472,6 @@ func (c *Manager) fallbackKill() error {
 }
 
 func (c *Manager) Delete() error {
-	var (
-		tasks    []uint64
-		threaded bool
-	)
 	// Kernel prevents cgroups with running process from being removed,
 	// check the tree is empty.
 	//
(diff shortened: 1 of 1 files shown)


cgroup2/manager.go
```diff
@@ -485,13 +481,13 @@ func (c *Manager) Delete() error {
...
 	}
 
-	if threaded {
+	var tasks []uint64
+	switch cgType {
+	case Threaded, DomainThreaded:
 		tasks, err = c.Threads(true)
-	} else {
+	default:
 		tasks, err = c.Procs(true)
```
thaJeztah, line 491: Alternative could be to just to try  either if we get anything in return?

```go
if tasks, _ := c.Threads(true); len(tasks) > 0 {
	return fmt.Errorf("cgroups: unable to remove path %q: still contains running tasks", c.path)
}
if tasks, _ := c.Procs(true); len(tasks) > 0 {
	return fmt.Errorf("cgroups: unable to remove path %q: still contains running tasks", c.path)
}
```

thaJeztah: oh, not sure if that makes sense. Looking at `Procs` and `Threads`, they both call `getTasks` under the hood; if the above _does_ make sense, I guess that function could possibly accept  both `cgroupProcs`  and `cgroupThreads` to get "any".

dcantah (approved):

fuweid (approved):

PR #361, opened 2025-03-19
support network metrics for cgroupv2

zouyee:
xref: https://github.com/containerd/cgroups/pull/81


cgroup2/stats/metrics.proto
```diff
@@ -120,3 +121,15 @@ message HugeTlbStat {
 	uint64 max = 2;
 	string pagesize = 3;
 }
+
+message NetworkStat {
+	string name = 1;
```
akhilerm, line 126: Is this the interface name?

akhilerm (commented):
I am thinking if I can use this in https://github.com/containerd/containerd/pull/10691 for getting the network metrics rather than via https://github.com/vishvananda/netlink that the CRI-O implementation uses for network metrics

cc: @mikebrow

cgroup2/stats/metrics.proto
zouyee: yes

zouyee:
> I am thinking if I can use this in [containerd/containerd#10691](https://github.com/containerd/containerd/pull/10691) for getting the network metrics rather than via https://github.com/vishvananda/netlink that the CRI-O implementation uses for network metrics
> 
> cc: @mikebrow

https://github.com/opencontainers/runc/blob/main/libcontainer/container_linux.go#L152

zouyee:
@dcantah @mikebrow PTAL

dims (approved):

mikebrow (approved):
LGTM

mikebrow:
> I am thinking if I can use this in [containerd/containerd#10691](https://github.com/containerd/containerd/pull/10691) for getting the network metrics rather than via https://github.com/vishvananda/netlink that the CRI-O implementation uses for network metrics
> 
> cc: @mikebrow

Will have to review the delta between using the NetworkStats returned from libcontainer through runc with the vishvananda/netlink code base.. probably do a review?  @MikeZappa87 thoughts?

MikeZappa87:
@mikebrow I am not sure how this repo fits together. Is this between the shim and the containerd?

mikebrow:
> @mikebrow I am not sure how this repo fits together. Is this between the shim and the containerd?

Yes this is code/repo defines cgroups 1/2 and we use it in containerd when interacting with runtime engines through the shim.  Eg. would be used in pulling in metric we are monitoring and conversion.. well it doesn't yet work for this new type but does for pid/cpu/memory... we're going to need to pipe the netstats through and thus begs the question which metrics are there differences.. needs investigation.

PR #369, opened 2025-06-15
Add cgroup2 hugetlb failcnt metric

gavinkflam:
We are missing hugetlb failcnt metric for cgroup2.

According to [kernel doc][1], this can be sourced from `hugetlb.<hugepagesize>.events`.

> hugetlb.<hugepagesize>.events
> A read-only flat-keyed file which exists on non-root cgroups.
> 
> max
> The number of allocation failure due to HugeTLB limit

[1]: https://docs.kernel.org/admin-guide/cgroup-v2.html#hugetlb


gavinkflam:
@AkihiroSuda Noticed you recently reviewed a similar PR. Would you mind taking a look at this one as well when you have a chance?

akhilerm (approved):

